PR #168, opened 2013-10-12
Bugfix for SDCard on AT90USB1286

Jerome-PS:
This should fix SDcard access on Printrboard and Teensly.


repetier:
Using new sdfat lib now.

PR #266, opened 2014-03-03
Edit with atmel studio

kyrreaa:
Added example project files for Atmel Studio with examples of paths to libraries for Arduino included in the projects.

The SD8495 and the AD595 can not share code and it was a bug in earlier code that made it appear so.
Missing break was in the old branch.


repetier:
Now we have PlatformIO integration making it easier to develop.

PR #334, opened 2014-11-04
delta distortion correction, autoleveling fixed, disable automatic fan, show zprobe

josefpavlik:
hi
I made a couple of fixes and features.
1. Zprobe state is visualized on display when you move Z axis from UI
2. Fixed autoleveling G29 and G32
3. Now you can disable automatic fan (autostart on 50 celsius) 
4. new feature - distortion correction for delta printers
This last feature measures grid of points (21 points measured plus 4 extrapolated) and compensates the Z coordinate using this matrix.
The delta printer needs this compensation due to non linear distortion. Distortion of other axes are not corrected as there is not possible to measure them. Works with G29 command. It's compatible with Marlin firmware, however the code is remade from scratch.

see the commits for detailed description


josefpavlik:
I've rewrote the distortion correction in integers and I've moved the correction to the function transformCartesianStepsToDeltaSteps
Now it's faster and consumes few memory. 

https://github.com/josefpavlik/Repetier-Firmware/commit/585278efa6471b6d92eaf27c7d1eec793562ce2f

repetier:
Just as a double notice. It is already included in 0.92 with integer computation and how I think it should be implemented. Works great for me on a arm.

PR #372, opened 2015-03-05
Some changes on Printer::buildTransformationMatrix

Jantek:
Hi,

I found some multiplies with zero witch I commented out. I am not really sure, but shouldn't you also use autolevelTransformation[1] to calculate the length of the first column vector? Shouldn't autolevelTransformation[1] also get divided by that length?

Kind regards,
Jan Phillip Willmann

PS: Your Software is GREAT WORK!!!


repetier:
I think you are right that 1 should be included. After all we build here rotation matrix with determinant 1 and ignoring one possible length can skew matrix. So I included it and hope it is ok:-) Just can't imagine how this could be unseen for so long.

PR #373, opened 2015-03-05
Changed the internal precision of the transformation matrix for the DUE.

Jantek:
Hi again,

it may have no great influence but I changed the precision for the calculation of the transformation matrix for the DUE from float to double (Lots of multiplications/divisions and square root operations with small numbers). Maybe it is a bit paranoid but the DUE should got enough memory for that so it shouldn't harm.

Kind regards,
Jan Phillip Willmann


repetier:
Sorry, but I don't think that is needed.
1. It must be identical to the AVR version as it is not hardware related and AVR does not know double
2. It still would get stored as float, so no advantage here.
3. I think precision is high enough even with float.

PR #377, opened 2015-03-28
Fixed Alligator board SPI EEPROM 

MarcoAntonini:


repetier:
Thanks for the update.

Needed to make some changes to your condition in Printer.cpp and SdFat as tehse files are identical for AVR and Due and AVR has no eeprom type. So I added defined() in addition.

Also put it in online config tool with new primary port selector, so you can now select programming or native port from config tool directly. No more tricks required.

PR #382, opened 2015-04-05
Fixed MICROSTEPPING16  for Alligator Board

MarcoAntonini:
Hi Roland,

in Repetier.h,  the cotrol :

```
#if (MOTHERBOARD == 501) 
```

not work, need to include Configuration.h file first.
Fixed some columns in microstepReading().

Thanks
Marco


repetier:
Ok, have changed it now.

Your position if the config include was not choosen well as you now have not some definitions needed in the config. I have instead moved the microstepping definition below the include.

MarcoAntonini:
Thanks :)

PR #383, opened 2015-04-06
Added bluetooth support for Melzi

NeoTheFox:
So this is some kind of bug, as seen in pins.h


repetier:
Have included the fix and added some more devices to the possible list.

PR #384, opened 2015-04-06
Fix u8glib block UART0, add Alligator support for RepRapDiscount Full Graphic Smart Controller

MarcoAntonini:
Hi Roland,

I added the  pinout for supporting RepRapDiscount Full Graphic Smart Controller on Alligator board.
Alligator has a software switch to select the voltage of the logic for the display, I added UI_VOLTAGE_LEVEL in Configuration.h.

When the display initializes I found that u8glib write the pin 0,  in many board pin 0 corresponds to RX pin of UART0, it causes a block of reception of the uart !
I fixed it by setting U8G_PI_CS2 to U8G_PIN_NONE 

Thanks
Marco


repetier:
I have added your changes with a little twist. Level selection is inside ui.h controller select. No need to put it in config also you can still override it in original config.

BTW: How do you connect the GLCD - do you have a adaptor for it as the pins have no 2 blocks as I would need to connect natively like with RUMBA.

MarcoAntonini:
Thanks.
The display need an adapter and connects on the expansion port (near to the sd card), i will give you one of this adapter.

repetier:
Thanks. On Rumba and RAMPS the same adaptor also works for their 4x20 line displays. Will yours also work or is this not possible with the pin layout. I know the GLCD uses SPI and not 7 pins like 4x20 displays but if it works that would be great as an options for all users.

MarcoAntonini:
Thanks for your advice, unfortunately on the expansion connector there are not enough pins for 4x20 display.
I'm thinking of an adapter that includes expansion connector display + PWM expansion connector, in this way I should have all the pins needed.

Excellent solution might be to use the RaspberryPi expansion port  with the new Repetier-Server and a nice touch color display connected to Raspberry, in the next few days I will try this configuration :)

repetier:
Next server release (0.60) will have a special frontend for touch screens on printers. It will be much simpler then the normal web frontend, so it can run on a Raspberry on HDMI port + touchscreen or over extension port.

PR #388, opened 2015-04-08
Work092

HakanBastedt:
Sync with upstream


repetier:
Can you explain what this FSR pin is. It does not seem to be the signaling pin for programming something. Also do all FSR sensors have and need this? Do they forget settings? I always thought you have some electronic to make them switch at a force you want and that is all. So your solution seems very specific and I'd like to know this before I include such code. After all I'm the one getting all the questions and I do not even own a FSR.

HakanBastedt:
Sorry,
it was not the intention to forward the code to you.
Please discard it. I am really a beginner at Git and github.

This is my own solution for Force Sensitive Resistor, with a small
Arduino board involved.The FSR_PIN is used to tell the
Arduino-board to find a Zero level just before the measurements.
I have found that temperature and other things makes the
FSR value drift quite a bit and it is necessary
to find a zero level just before measurement.

I feel a little embarrased to use git and bithub in the wrong direction.
Please throw it away.

Thanks
Hakan


2015-04-09 17:45 GMT+02:00 repetier <notifications@github.com>:

repetier:
ok, no problem.

PR #395, opened 2015-05-02
Additional fastio.h pins for Alligator board

MarcoAntonini:
Hi Roland,

Alligator uses more pins than ArduinoDUE, I mapped these pin (PIN-92-100) in Fastio.h to resolve compile errors with Alligator board.

Thanks

### DIFF
--- a/src/ArduinoDUE/Repetier/fastio.h
+++ b/src/ArduinoDUE/Repetier/fastio.h
@@ -270,6 +270,37 @@
 #define DIO91_PORT PIOB
 #define DIO91_PIN PIO_PB15A_CANRX1|PIO_PB14A_CANTX1
 
+// Additional Pins for Alligator board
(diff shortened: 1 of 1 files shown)


repetier:
Thanks for the update. Have you seen I also updated variants.cpp in additional files to work with Arduino IDE 1.6.x ?

Is there a reason for not enabling the watchdog? You know for arm you can only set watchdog once. Maybe we should remove that cod epart completely from variants and move it to HAL::hwInit so user can decide.

MarcoAntonini:
watchdog was just an oversight, better to choose on HAL.
Comment please WDT_Disable (WDT) in variant.cpp .

Thanks

repetier:
Ok will do that. Have just added the disable locally so it is not needed in variants anyway.

MarcoAntonini:
Ok Thanks

PR #397, opened 2015-05-13
Add MCP4728 DAC (Digital Trimpot) support for PrintrBoard Rev. F

not404:
Creates a new motherboard type (92) for the PrintrBoard Rev. F., which uses an MCP4728 DAC as a digital trimpot. 

Implements additional M-codes as PrintrBot's fork of Marlin does. (M907, M908, M909, and M910) - PrintrBot's fork of Marlin was used as a guide or reference implementation.

### DIFF
--- a/src/ArduinoAVR/Repetier/Commands.cpp
+++ b/src/ArduinoAVR/Repetier/Commands.cpp
@@ -416,6 +416,129 @@ void motorCurrentControlInit() //Initialize Motor Current
 }
 #endif
 
+
+#if STEPPER_CURRENT_CONTROL == CURRENT_CONTROL_MCP4728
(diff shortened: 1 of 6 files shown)


repetier:
Nice commit. I have added percent settings for other boards as well and made it the default method. Guess percents work better across different current value ranges.

not404:
Thanks, glad these additions have proven to be useful for some of the other boards as well.

About the only thing that seemed "iffy" with the Trimpots to me is the fact that there's no guard to enforce a maximum value on setMotorCurrent() or setMotorCurrentPercent().

I'm OK with that for my own needs (I may want to drive the stepper chip at 125% for brief periods of abuse testing, for example), but to save lesser-experienced end users from themselves, adding a guard to enforce a maximum limit on each setMotorCurrent() function is probably worth doing.

repetier:
You wrote that ungarded part :-) For next update I have added a guard so that values > 100% are clipped to 100%.

PR #399, opened 2015-05-15
Fix Alligator MOTOR_CURRENT

MarcoAntonini:
Hi Roland,
good idea to set the current with a percentage, I need a small change for Alligator to fix compiler error .

Thanks


repetier:
Thanks for the check. Stupid error from my side. Had the same for an other board as well:-(

PR #401, opened 2015-05-17
Add support for Makers Tool Works Led Strip

exuvo:
Implemented using the new event system.
New config option CUSTOM_EVENTS_MTWLED, requires CUSTOM_EVENTS.
Also includes fix for incorrect CustomEventsImpl.h include in Printer.cpp .

The only thing I'm unsure about is a call to HAL::i2cInit(400000L) which might conflict with i2c speed settings used by potential UI panels.

Everything compiles in Code::Blocks (after .ino rename to .cpp) and works on my MendelMax 3 with a RAMBo 1.3L.

### DIFF
--- a/src/ArduinoAVR/Repetier/Configuration.h
+++ b/src/ArduinoAVR/Repetier/Configuration.h
@@ -334,7 +334,7 @@ The codes are only executed for multiple extruder when changing the extruder. */
 #define EXT0_EXTRUDER_COOLER_PIN -1
 /** PWM speed for the cooler fan. 0=off 255=full speed */
 #define EXT0_EXTRUDER_COOLER_SPEED 255
-/** Time in ms between a heater action and test of success. Must be more then time between turning heater on and first temp. rise! 
+/** Time in ms between a heater action and test of success. Must be more then time between turning heater on and first temp. rise!
  * 0 will disable decoupling test */
 #define EXT0_DECOUPLE_TEST_PERIOD 18000
 /** Pin which toggles regualrly during extrusion allowing jam control. -1 = disabled */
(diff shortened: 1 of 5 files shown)


repetier:
That was a quick response. Haven't even had time to write a own version.

Only problem is that the idea is not to have CustomEvent implementation in main source code, as each printer(combination) might need different solutions here. On the other side is it great to have working examples. So what I did is make a subfolder SampleEventSystems where we can collect multiple implementations if they are likely to spread. So I put your code minus the defined test there. So users wanting it can simply copy it.

exuvo:
Thanks. It feels good to have contributed something worthwhile. I sent a message to MTW suggesting that they add text to their store front that repetier firmware has support for the LEDs.

In the configurator: Do you think it would be a good idea to add a selection field on the User Interface tab similar to the current Display Controller selection for custom event systems. It would need to copy all files from the selected Sample directory and add the CUSTOM_EVENTS define to the config. Probably needs a disclaimer that further configuration might be necessary. I would send a pull request for that but I can't seem to find any repository for the configurator.

Regarding the RAMBo pin addition, what was the reason for not accepting?

repetier:
I have added the pins later on but I can not find a reference page for RAMBO saying which pin has which function. I saw it supports 2 extruders + bed and am a bit surprised about the 4th thermistor. Even marlin seems not to know it is there.

Regarding config tool - I'm thinking about it. Will need a special solution for it as you need extra files, so it is not that simple. Especially if you only download the config file you have a problem. I guess I will make a selector for the directories and issue a warning in the download page that you need to download at least once the zip which then will insert the files. Once that is set up i can simply add new directories for other printers/solutions.

Maybe I also add a readme.txt per directory so every solution can explain what it does and which configurations are needed. That would help users with setup and with the new manual config insertion we can already insert any additional setting.

exuvo:
I got the pinout from the top left corner of the [schematic](http://reprap.org/mediawiki/images/9/91/Rambo1-2-schematic.png).

4 of the N-MOSFET outputs are identical and are named Heat 0, Fan 0, Heat 1, Bed. Then there is two weaker ones for Fan 1 and Fan 2. You could run a third extruder on Fan 0 if you add an external motor driver for it. And then you have a use for the 4th termistor connection.

Send me a message if you want me to write up a config readme for the MTW LED Strip.

I also checked with the original LED Strip author, OhmEye, and he was fine with my port.

repetier:
I'm not really sure that is correct:

#define TEMP_3_PIN     7

if I check the schematic 7 = AD7 is X_EN and not a thermistor. On the other side did I fail to understand what names are used for thermistors. Not really my field.

exuvo:
The ADC is on port F and K. You are looking at AD which is for external memory. You should be looking at pin PF7 for THERM3.

repetier:
I see, thanks for the info. I was mislead by the AD at the top. Now I see it, it is obvious:-)

PR #404, opened 2015-06-02
Safer homing algorithm for deltas

kyrreaa:
Safer homing algorithm that ensures homing switches respond as expected during homing.

By checking the endstops status after each step in the homing sequence we can catch any sticky or faulty endstops.
This ensures that we do not blindly assume we are homed just because the routine finishes.
Avoids head crashes and potential damage/fire...

To test it, simply hold any (X,Y or Z tower) endstop when it is not supposed to, or prevent an axis from triggering their endstop when they try.


repetier:
I think i is a good test, also I know I will now get an error on my test boards as these rely on success on error having no printer connected :-)

PR #409, opened 2015-06-20
WIP: Allow homing of dual Y steppers individually

javawizard:
(This is still a work in progress - hasn't yet been tested, and there are a few things that really ought to be cleaned up. Posting this now so that I can get some feedback on the direction I'm going, and particularly switching endstop statuses and direction flags to `flag16_t` as opposed to, say, having separate fields for the statuses of the different endstops and which motors have been killed. I'd also like to generalize this to X and Z as well - Z would be particularly useful with the plethora of printers with two separate steppers to raise the gantry.)

When using FEATURE_TWO_YSTEPPER, a second endstop can be set up
by setting MIN_HARDWARE_ENDSTOP_Y2 and Y2_MIN_PIN. When these are
set, hitting the first Y endstop causes only the first stepper to
stop. The second endstop must be hit in order for the second
stepper to stop as well. This allows both sides of a Y gantry to
be synced up automatically.

### DIFF
--- a/src/ArduinoAVR/Repetier/HAL.h
+++ b/src/ArduinoAVR/Repetier/HAL.h
@@ -143,6 +143,7 @@ typedef uint16_t speed_t;
 typedef uint32_t ticks_t;
 typedef uint32_t millis_t;
 typedef uint8_t flag8_t;
+typedef uint16_t flag16_t;
 typedef int8_t fast8_t;
 typedef uint8_t ufast8_t;
 
--- a/src/ArduinoAVR/Repetier/Printer.cpp
+++ b/src/ArduinoAVR/Repetier/Printer.cpp
@@ -159,11 +159,11 @@ int debugWaitLoop = 0;
 fast8_t Printer::wizardStackPos;
 wizardVar Printer::wizardStack[WIZARD_STACK_SIZE];
 
-flag8_t Endstops::lastState = 0;
-flag8_t Endstops::lastRead = 0;
+endstops_t Endstops::lastState = 0;
+endstops_t Endstops::lastRead = 0;
 
 void Endstops::update() {
-    flag8_t newRead = 0;
+    endstops_t newRead = 0;
 #if (X_MIN_PIN > -1) && MIN_HARDWARE_ENDSTOP_X
         if(READ(X_MIN_PIN) != ENDSTOP_X_MIN_INVERTING)
             newRead |= ENDSTOP_X_MIN_ID;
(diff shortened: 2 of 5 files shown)


javawizard:
Also thinking it'd be nice to be able to set a limit on how many more steps one side can continue running after the other hits an endstop before the printer concludes that something's terribly wrong and yells at the user to manually sort things out.

repetier:
Ok, here my first thoughts. Adding additional endstops is something needed for very special printers. Most common would be for z. I think for z we even need 3 endstops tripple motor handling. That would cover all printers moving with 3 spindles.

Problems with your solution:
- Very special - only y2 min. I even do not understand when that could be needed.
- Has a big speed impact for all other printer especially on 8 bit computer. You need to extend dir (which I could not see you did) and endstop hit flags to 16 bit more then doubling all operations on it and they are all in teh stepper interrupt which is  time critical.
- Increased memory solution for deltas which are very critical at least.
- Neglecting the due port if you make HAL changes.

So first thing would be reduce overhead to printers that require that much end stops. So for endstop flags I would suggest 8 or 16 bit variable depending on EXTENDED_ENDSTOP_SUPPORT which would be enabled if any xyz2 or 3 endstop gets defined.
For extended endstop not using dir to detect triggered endstop but a new variable waitingEndstops with x1,x2,y1,y2,z1,z2,z3 as flags. I will add a third motor mirror then. This will still allow fast usage of dir in all non endstop related parts which are quite a few.
The second endstop will normally trigger same time as first endstop, only during homing it might be different, which is why you go all the pains, right? So it would be a idea to introduce a homing/zprobing flag to switch between testing normal endstops and additionaly endstops. Printer::flag2 has free bits so lets take 64 for PRINTER_FLAG2_HOMING

Regarding your concern with rest move length I think you are right. Motors should not go apart too much, so limiting it to a few mm would be a good idea. So if we see homing flag and first one triggers we could limit stepsRemaining to length*stepspermm if homing flag is enabled.

If your modification goes into this direction I will be gladly pull your changes. If you have any other questions or remarks on my remarks let me know. These are just the first things I would change to save existing users from new problems.

javawizard:
Cool, exactly what I was looking for.

* I definitely anticipate Z being the biggest use case for this. Ability to support arbitrary numbers of motors and endstops would be cool as well, although I imagine that'd take a sizable bit more work than just adding support for a second endstop to match the second motor supported with `FEATURE_TWO_?STEPPER`.
* Haha, yeah, that was the last bug I had to work out - I meant to change dir to `flag16_t` but missed it. I'll post that shortly. So use a separate variable for endstop statuses and storing which motors have been killed then? That sounds good to me if so.
* I'll post a picture of my printer once it's done so that you can see why I'm using this for Y specifically.

Thanks for the feedback!

repetier:
I only want to extend z to 3 axes whcih would then get
FEATURE_TWO_ZSTEPPER and FEATURE_THREE_ZSTEPPER

As with the kill bits - I would take them as direction independent so we only need one extra byte which also does not need to be in printlin eor as a static variable to safe ram. Since we do this only during homing that axis will not move until the homing move is finished and can then be reset to 0.

javawizard:
That works as well, although that would necessitate an additional bit somewhere to distinguish whether a given move is a homing move, which (as far as I was able to tell) doesn't currently exist.

repetier:
Yes, as I wrote:

Printer::flag2 has free bits so lets take 64 for PRINTER_FLAG2_HOMING

so I would add that for homing moves and also reset the kill bits at that position then, so we always start with 0 bits.

javawizard:
Ah, missed that. Yep, sounds good to me.

Kulitorum:
A friend of mine just implemented individual endstops for 4 steppers on z-axis in a very clever way:

He hooks 4 drivers to the same output on the board, using just the step/direction signals.
Each driver has the "direction" pin running through that motors endstop, so when the motor reaches the endstop, the direction is reversed for just that one motor, and it will effectivly wait for the other endstops, because the motor that is triggered will then return one step. The endstop will be disengaged, and the motor will move down again. This keeps happening in a loop, until all endstops are triggered at once, and the homeing stops.

So simple, so effective :)

PR #413, opened 2015-07-17
Added support for Alligator Piggy multi-extruder expansion

MarcoAntonini:
Hi Roland,

Alligator Piggy expansion module is ready, I added support for a second DAC and updated pinout of Alligator Board up to 4 extruder. 
I will send you a Piggy module for testing.

thanks
Marco


repetier:
Thanks for the piggy module.

Changes are added. Please be careful if you change chared files like commands.cpp. The avr is the original and with each update I copy the avr version to the due. So making changes only to due will loose the changes if I do not see it in time. Today I saw it and copied it to avr, so no damage here, but I might some day not see it:-)

MarcoAntonini:
I always forget original avr,  next time I'll be more careful :)

PR #414, opened 2015-07-23
Update SDCard.cpp

justM4D:
the kill-command should have false as parameter, as it's supposed both heaters and steppers, and not only steppers, as far as I understand it


repetier:
Yes, you are right.

jamesarm97:
Yes. I have been ignoring this one long enough. Every time I stop an SD print it would leave the heaters going and I would have to stop them. Thank you.

PR #418, opened 2015-08-11
Watchdog enabled before sd init

MarcoAntonini:
Hi Roland,

In SAM processor during Powerup if the SD card is inserted cause a watchdog reset.
I solved by enabling watchdog before init sdcard.


repetier:
That might even make sense if we assume that on first power up the arm is initialized with a short watchdog timer. After that it may keep the longer timing we set.

PR #425, opened 2015-09-03
Arduino Due native usb

Salandora:
This is a fix to get the native USB port of an Arduino Due to run properbly.
Short story:
PWM interrupt routine conflicted with the native USB port, causing the printer to stop working.

Long story:
After I got my RADDS board I switched to the Repetier Firmware, of course I wanted to use the native USB port for communication to not have my prints stutter or anything because of slow Serial communication.

So I generated my Firmware with SerialUSB support. 
I tried to print something and had to notice that the Print stopped after a few seconds.

After a few monthes now I had time to investigate the problem. I firstly wanted to know what is causing it and to not waste material I switched to dry run mode.
But in dry run mode the problem didn't appear.
So I had the guess It must have something to do with the extruder or the temperature handling.

So I investigated it further and noticed that there was a InterruptBlock class introduced.
And the HAL interrupt functions got disabled.

I found that the PWM interrupt was using the InterruptBlock class to prevent interrupts.
And I found that at the bottom Interrupt's should have been allowed again (HAL::allowInterrupts()).
But of course it couldn't enable the interrupts because the function got disabled.
So I changed the HAL function to the unprotect function of the InterruptBlock class and reflashed the firmware to the Arduino.

After a bit of testing I didn't got any stops anymore.
So I guess this fixed the problem for me and maybe for others too.

Greetings
Salandora

### DIFF
--- a/src/ArduinoDUE/Repetier/HAL.cpp
+++ b/src/ArduinoDUE/Repetier/HAL.cpp
@@ -1102,7 +1102,8 @@ void PWM_TIMER_VECTOR ()
   if (pwm_pos_set[NUM_EXTRUDER] == pwm_count_heater && pwm_pos_set[NUM_EXTRUDER] != HEATER_PWM_MASK) WRITE(HEATED_BED_HEATER_PIN, HEATER_PINS_INVERTED);
 #endif
 #endif
-  HAL::allowInterrupts();
+  //HAL::allowInterrupts();
+  noInt.unprotect();
   counterPeriodical++; // Appxoimate a 100ms timer
   if (counterPeriodical >= 390) //  (int)(F_CPU/40960))
   {


repetier:
Interesting analysis. You are right that HAL::allowInterrupts() is a empty function and your fix is at least the right way to allow interrupts again, which was at least the intention.

I'm normally using programming port so I never noticed this. In short tests it always worked. And I'm still not sure why that should make a difference. The only thing I can think of is timing. If temp. interrupt is blocking too long it might give problems with usb timing/interrupt. Anyhow I have included it for next release and will test now with native port. So thanks for the fix.

Salandora:
Thanks for the feedback, I also guess it has to do with the timing.
In the bottom part of the Timer interrupt there is some Analog input processing wich might take to long.

I also saw that the comment said read analog values -- _"only read one per interrupt"_
But If I'm correct then all Analog Inputs get's processed.
Is the comment wrong now or the code or is it me.

I also found a few more HAL::allowInterrupts() and HAL::forbidInterrupts() calls which are now doing nothing. Are they supposed to become converted to InterruptBlocks?

kyrreaa:
What would be nicer is really if the pwm for heaters could run from hardware properly.
If a pwm capable pin is selected as io then this should be default behaviour.

Anyway, good catch!

Kyrre

jamesarm97:
This has been a debate brewing recently by a few in the groups I hang out in. Seems there is some potential safety issues with software PWN and board lockups leaving the heaters enabled. 

--
James Armstrong

repetier:
@jamesarm97 Good hint about safety. At least with watchdog enabled you are right I think.

@kyrreaa No will never again go back to hardware pwm. That is how I started and it was hell. Every board if it uses pwm pins at all uses different timers and we use several timers for internal uses, so these also must be different from the hardware pwm. So finding a working combination for all boards is a nightmare to program. At some point I threw it all out and made everything software pwm.

@Salandora HAL::allowInterrupts() and HAL::forbidInterrupts() are meant to be used inside interrupts and are not empty for AVR boards. On due I do not use them since it gives problems with all the different interrupt priorities. Having them in bresenham with function caused frequent problems but due was fast enough to allow without reenabling. So having some called from here is for the avr solution.

InterruptBlock is a very hard block that even forbids higher priority interrupts (which might be the problem here), so we should not overuse them inside interrupts which are the only source of a change and need no further protection.

kyrreaa:
@repetier
From a professional standpoint, when you have a hardware unit designed to do a job you better have a good reason not to use it.
Since board specific files should handle the timer stuff I don’t think that is a very good reason not to use it.
Then again, I don’t like to use any arduino stuff either. I’d rather have the hal do it.
I can understand the frustration and the throwing it out approach though, but seriously 15hz synchronized pwm for heaters is very very very bad.
It increases the supply requirements and noise a lot. Most printers I have looked at do not meet emi requirements and are actually a bit risky to sell as the seller is liable for any violated regulations.
PWM’s being locked on is also bad...

Btw, the pwm and queue stuff is probably responsible for the jerky motion on the Due platform with high rez, high speed deltas.
With heater on the motion for a fast circle in x/y plane is jerky but with heater off it is smooth.
Sometimes after long prints the test is also jerky, and resetting the board before running code again results in smooth motion.
(Haven’t committed any report on this yet as I am trying to pinpoint what is happening first...)

rcarlyle:
@repetier 
If the watchdog timer works properly, software PWM is ok. But running software PWM without a watchdog is a completely unacceptable programming practice. If the board freezes while the heaters are on, software PWM will output 100% power. Whereas hardware PWM maintains the last power output (which is typically under 50%). Most people's hot ends will melt if 100% power is locked on. That's a serious fire hazard. 

Calling the watchdog reset in non-heater code prone to going into lock-up (like SD card reads) is another very, very bad programming practice. You don't want the cause of the lock-up to be continuously resetting the timer. That defeats the point. 

If you're going to rely on software PWM, the watchdog should be enabled by default and only reset by the heater code.

repetier:
Best placement for watchdog is a hard decision. Heater code sounds good and that's where it is (except some sd stuff that can otherwise cause watchdog). Problem is what we want to catch. Heater code are 2 functions.
1. Main thread managing set PWM to adjust to measured temperature.
2. PWM interrupt enabling and disabling the heaters.

If we have it in 1. we know that the main thread is running and assume that interrupts still get called frequently. Having it in 2. we could emit sd watchdog. Timer should be called frequently and we do not care about main thread. So a deadlock in main will not trigger watchdog but we keep pwm on. But that gives me an idea. 

How about main thread setting a flag and pwm interrupt triggering watchdog only with set flag. That way both must be called frequently to keep it up and running. Then sd watchdog is also no problem if it is only internal flag.

Regarding PWM interrupt and protection I think it might even not require a protection at all. I'm just doing test print without protection. That should also solve any timing issues with stepper becoming jerky since that interrupt has higher priority, also I do not see that it could influence much. Never felt my high resolution delta was jerky with heaters on. Might delay a step here and there, yes. But may depend on your set frequency. On the other side pwm for extruder gets also called for disabled extruders, so why should that make a difference?

Salandora:
I do have to tell you that my fix did not fixed my problem to 100%.
I tried to print yesterday and in middle of the print it suddenly stopped again. But only communication. 
Heater managing and all went on normally. Checked this by looking at the LED's which indicates the status of the mosfets. 
Before this fix I already tried 3 different usb cables ao I guess this is not the problem. 
Need to further investigate this which could become a problem.
Any ideas?

repetier:
I'm currently also printing over native port and get a hang every 30-50 minutes or so.  Very strange thing. Firmware works as it should and server is also running. After about 2 minutes print continues like nothing happened, only that server got some time outs of course but sending new commands did not work. Looks like the usb communication has a complete hang and at some time starts working again and then it continuous working like before. Just trying to let it happen again and see if log reveals something.

repetier:
Ok, it happened again. Here the log

11:21:41.597: ok 35267
11:21:41.599: N100807 G1 X14.266 Y-8.698 E7.72056
11:21:41.600: ok 35268
11:22:12.908: Warning: Communication timeout - resetting communication buffer.
11:22:12.909: N100808 M105
11:22:12.919: Warning: Communication timeout - resetting communication buffer.
11:22:12.920: N100809 M105
11:22:12.921: N100810 G1 X14.832 Y-8.306 E7.72827
11:22:12.932: N100811 G1 X15.574 Y-7.756 E7.73861
11:22:44.934: Warning: Communication timeout - resetting communication buffer.
11:22:44.935: N100812 M105
11:22:44.946: Warning: Communication timeout - resetting communication buffer.
11:22:44.946: N100813 M105
11:22:44.947: N100814 G1 X16.652 Y-6.877 E7.75418
11:22:44.957: N100815 G1 X17.676 Y-5.942 E7.76970
11:23:16.955: Warning: Communication timeout - resetting communication buffer.
11:23:16.955: N100816 M105
11:23:16.966: Warning: Communication timeout - resetting communication buffer.
11:23:16.966: N100817 M105
11:23:16.967: N100818 G1 X18.690 Y-4.903 E7.78595
11:23:16.978: N100819 G1 X19.703 Y-3.733 E7.80328
11:23:23.882: ok 35269
11:23:23.903: ok 35270
11:23:24.050: ok 35271
11:23:24.072: ok 35272
11:23:24.073: T:214.74 /215 B:50.03 /50 B@:0 @:220 T0:214.74 /215 @0:220 T1:37.52 /0 @1:0
11:23:24.074: ok 35273
11:23:24.075: T:214.74 /215 B:50.03 /50 B@:0 @:220 T0:214.74 /215 @0:220 T1:37.52 /0 @1:0
11:23:24.076: ok 35274
11:23:24.175: ok 35275
11:23:24.191: ok 35276
11:23:24.192: T:214.39 /215 B:49.91 /50 B@:255 @:220 T0:214.39 /215 @0:220 T1:37.43 /0 @1:0
11:23:24.193: ok 35277
11:23:24.195: T:214.39 /215 B:49.91 /50 B@:255 @:220 T0:214.39 /215 @0:220 T1:37.43 /0 @1:0
11:23:24.196: ok 35278
11:23:24.199: ok 35279
11:23:24.207: ok 35280
11:23:24.208: T:214.39 /215 B:49.91 /50 B@:255 @:220 T0:214.39 /215 @0:220 T1:37.43 /0 @1:0
11:23:24.208: N100820 M105
11:23:24.208: ok 35281
11:23:24.211: T:214.39 /215 B:49.91 /50 B@:255 @:220 T0:214.39 /215 @0:220 T1:37.43 /0 @1:0
11:23:24.212: N100821 G1 X20.541 Y-2.640 E7.81869
11:23:24.212: ok 35282
11:23:24.214: N100822 G1 X21.318 Y-1.504 E7.83410
11:23:24.215: ok 35283
11:23:24.218: N100823 G1 X21.987 Y-0.389 E7.84866
11:23:24.218: ok 35284

What is interesting here is that it continues with a lot of ok. Even
11:22:12.909: N100808 M105
is acked by 
11:23:24.072: ok 35272

So no line send got lost and no line received got lost on both sides. Just like if the usb package did take a bit longer then normally. So your conclusion that something is blocking too long is not that wrong I guess. I have no idea how usb works but for me it looks like a package was not acked fast enough and some time later it was tried again. Something we surely want to prevent.

Salandora:
Yup same problem here.
I also switched the arduino usb implementation just to check if this is the problem. But that also didn't solve it.

I still guess it has something to do with the heater interrupt.
Because in dry run mode it works.

luc-github:
On Davinci printer which is a DUE board based  which only have the native port available, the issue is present for a while actually, but I did not thinked it could be linked to FW itself, so I did not raised it. .https://github.com/luc-github/Repetier-Firmware-0.92/issues/38,  a workaround is to set a big cache like 2024.
Not sure if it help to find the root cause.

repetier:
I have now locally reduced pwm interrupt to priority 15 and also optimized the temperature loop. Removed the inner loop of 10, so making that part much faster. Hope that one of this helps. Now I'm back to printing to see if I still get the problem.

repetier:
No that did not help. After 3 hours without problems I got 2 hangs in 30 minutes.

I'm also wondering why dry run should have a effect on this problem. The interrupts is not really shorter if dry run is enabled. There we measure analog inputs and set pwm. We do this as well in dry run mode, just that all pwm are 0 so never get enabled.

If it is a long breaking interrupt it should be our stepper interrupt. It runs with priority 1 and with dry run enabled it is a very small bit faster. I did not understand how usb serial works so I never figured out with which priority the interrupts are handled. Since higher priority interrupts lower priority I will try now with all other priorities also reduced. Just a shot in the dark.

Salandora:
@repetier yeah you're right it doesn't make sense that dry run has any effect on this. But atleast for me  it has an effect. 
So let's think the other way. 
What happens when you activate dry run.

 - The temperature's gets resettet to 0 => no PWM?
 - Extruder is deactivated

Is there anything else?

repetier:
No, seems pretty much all. Just does not step extruder and pwm all off, everything else keeps the same.

Could it be electronic noise that causes confusion that takes so long? Do not really believe this. It is also a signal that it always seems to take around 102 seconds. Knowing that times are 32 bit with 1 tick per hz we need 2^32/84000000 = 51.1 Seconds. So if we have a timer with prescale 2 it would take exactly these 102 seconds I see in log. So what I think is that a timer signal gets lost/missed somehow and only the next time it comes along communication continues. I had the same problem with my stepper timer that I set it in the past and had to endure a full roundup, just because something interrupted it invalidating the computation time which should be near future.

To narrow it down I will make a dry run print to see if it also disappears for me, so we can be more sure that there is a connection.

repetier:
I think I have a fix now. Timer 1 interrupt has a prescale of 2 and if it missed the timer takes exactly 102 seconds like I got. I think the problem was the other way around. The USB interrupt interrupted stepper timer interrupt while setting clock so it finished after clock was set and so it missed the timer. The reason it happens only with extruder is that it happens during retractions where speed*resolution causes a higher frequency leaving less time for usb interrupt. Here my fixed solution in HAL.cpp until it is in work092 

/** \brief Timer interrupt routine to drive the stepper motors.
*/
void TIMER1_COMPA_VECTOR ()
{
  // apparently have to read status register
  stepperChannel->TC_SR;
  stepperChannel->TC_RC = 1000000;
  uint32_t delay;
  if (PrintLine::hasLines())
  {
    delay = PrintLine::bresenhamStep();
  }
  else if (Printer::zBabystepsMissing != 0) {
    Printer::zBabystep();
    delay = Printer::interval;
  } else
  {
    if (waitRelax == 0)
    {
#if USE_ADVANCE
      if (Printer::advanceStepsSet)
      {
        Printer::extruderStepsNeeded -= Printer::advanceStepsSet;
#if ENABLE_QUADRATIC_ADVANCE
        Printer::advanceExecuted = 0;
#endif
        Printer::advanceStepsSet = 0;
      }
      if ((!Printer::extruderStepsNeeded) && (DISABLE_E))
        Extruder::disableCurrentExtruderMotor();
#else
      if (DISABLE_E) Extruder::disableCurrentExtruderMotor();
#endif
    }
    else waitRelax--;
    
    delay = 10000;
  }
    // convert old AVR timer delay value for SAM timers
  uint32_t timer_count = (delay * TIMER1_PRESCALE);
  if (timer_count < 210) // max. 200 khz timer frequency
    timer_count = 210;
  InterruptProtectedBlock noInt; // prevent interruption or we might get 102s delay
  if ( stepperChannel->TC_CV + STEPPERTIMER_EXIT_TICKS > timer_count) {
     stepperChannel->TC_RC = stepperChannel->TC_CV + STEPPERTIMER_EXIT_TICKS; // should end after exiting timer interrupt
    //stepperChannel->TC_CCR = TC_CCR_CLKEN | TC_CCR_SWTRG ;
  } else {
     stepperChannel->TC_RC = timer_count;
  }
}


At least I got no pause in 10 hours of printing.

Salandora:
@repetier I tested your fix for a few days now, I haven't had any stops anymore.

repetier:
Great so I assume this issue is fixed then, since I also had no timeouts.
It will be included in next work092 update. Just need some more testing with the next release since we changed/added a few thousand lines.

Cannonized:
I'm having an issue where I'm not seeing errors in the code or wait commands. The buffer drops to zero the printer stops then goes again fluctuates back and forth I've adjusted Hal.h //Interval / (32lhz/128) = seconds
#define watchdog_inerval to 2048u // 16sec (~16 Seconds max)

I've changed //#define serial_buffer_size 1024. To 2048 to match the watchdog interval. 

My Hal. Cpp doesn't look anything like the posted fix and I'm not 100% on if that will fix my issue. Of buffer being irratic  on prints over 3 hours. I'm also not quite 100% with arduino programming to know if it's just a copy and paste over what's existing in the Hal.cpp file.

repetier:
Watchdog and serial buffer size have no relation. I also have no idea why you change them. Assuming you are using latest version I do not think it is the problem we spoke about here. When you go to default settings, how would your log look like when it happens. That would give the most usable informations I think to start analysing.

Cannonized:
Thanks for the response. It's not showing any errors in the log. Everything prints OK for about 2hrs then the buffer starts fluctuating up and down and the print head stops then starts again. I've notice that it seems to coincide with what sounds like the fan cicking on and temperature fluctuations of only a couple degrees on the print heat. It's almost as if it gets on information overload and just caps out. If there is another log I'm to look for please let me know I've only been using repetier 2 weeks now. The quality so far is way better than xyz but if it can't do more than a 2hr print I'm going back to stock.


Sent from Yahoo Mail on Android 
 
  On Fri, Feb 5, 2016 at 2:55 AM, repetier<notifications@github.com> wrote:   
Watchdog and serial buffer size have no relation. I also have no idea why you change them. Assuming you are using latest version I do not think it is the problem we spoke about here. When you go to default settings, how would your log look like when it happens. That would give the most usable informations I think to start analysing.

repetier:
I know users printing a week long without trouble, so that is not a normal problem.

Even without errors the log is interesting as it would show where it slows down and maybe why as you said it runs out of buffer and waits before starting. If your fan points not at the nozzle but also at the heat block enabling it will cool down extruder a few degrees until the heat controller can catch up on this new influence.

One important thing I forgot is, what are you sending with? Software and version. I'm now privately only using our server as sender as this works very reliable and I can still use the host to connect and send data.

Cannonized:
I'm using repetier host v1.6.0 uploaded 0.92 from luck github arduino 1.6.r2 not using the USB core.cppEverything has been working fine as long as I bring less than 2 hours. Anything more than that and the buffer becomes erratic 32 -12 - 28 - 32- 17 - 0 (printer freezes a fee seconds) then goes again for a couple of lines then the buffer on the lcd goes nuts again. Nothing is showing up in the log. I hit kill print and think I erased the last log file as all it is showing now is m105*(##) I adjusted the interval timer and such as a grasp at straws from luc-github issue 51 or 38 I basically Googled repetier .92 DaVinci 1.0 buffer issues after 2+hrs.I'm really not sure what to do. I'm running 225 for extruder heat and 110 for bed temp 

Sent from Yahoo Mail on Android 
 
  On Fri, Feb 5, 2016 at 12:12 PM, repetier<notifications@github.com> wrote:   
I know users printing a week long without trouble, so that is not a normal problem.

Even without errors the log is interesting as it would show where it slows down and maybe why as you said it runs out of buffer and waits before starting. If your fan points not at the nozzle but also at the heat block enabling it will cool down extruder a few degrees until the heat controller can catch up on this new influence.

One important thing I forgot is, what are you sending with? Software and version. I'm now privately only using our server as sender as this works very reliable and I can still use the host to connect and send data.

luc-github:
@Cannonized look like you are using my fork of repetier for Davinci printer
If yes, the fix for due native is in sync branch not yet in master of my github and so you are reporting in wrong place

Cannonized:
Luc, I guess I was in the wrong place. This issue is driving me batty. I will look for the sync branch issue on your github.

PR #429, opened 2015-10-07
Compiler delint - Clean up a number of compiler warnings

ezrec:
This patch series cleans up a number of compiler warnings, and should allow Repetier to compile cleanly with -Wall -Werror.


repetier:
0.92 has greatly reduced warnings compared to 0.91.

We do not change 0.91 any more, all new changes go to new version so I close this.

PR #432, opened 2015-10-23
Motorized Auto Bed Leveling for printers with three separate motors for Z-Axis (e.g. sparkcube)

Glatzemann:
Further information about this pull request (in german language): http://forums.reprap.org/read.php?334,571452

Video of the calibration procedure: https://youtu.be/bDfcVljKg7E

If you confirm to do the motorized bed leveling using the G33 command, I will create a entry for it in the reprap wiki (G-Code page).

Where should I write a manual how and why to use this feature? RepRap Wiki?


repetier:
New 0.92.8 now also contains a similar solution. Since I refactored the bed leveling routines into BedLeveling.cpp and added some handy function it does not need your code. But I think it is nearly the same. Extrapolate to motors and iterate until change is small enough.

PR #437, opened 2015-11-17
JSON-formatted responses (M408, M20 S2, M36)

toxuin:
This patch adds JSON-formatted responses to M20 S2, M408 and M36.
This makes Repetier-Firmware work with ESP8266 Duet web interface, PanelDue and probably some other things.
Tested with RADDS and RAPMS 1.4.

### DIFF
--- a/src/ArduinoAVR/Repetier/Commands.cpp
+++ b/src/ArduinoAVR/Repetier/Commands.cpp
@@ -1555,7 +1555,7 @@ void Commands::processMCode(GCode *com)
         }
         break;
 #endif
-#if JSON_OUTPUT
+#if JSON_OUTPUT && SDSUPPORT
     case 36: // M36 JSON File Info
         if (com->hasString()) {
             sd.JSONFileInfo(com->text);
--- a/src/ArduinoAVR/Repetier/Printer.cpp
+++ b/src/ArduinoAVR/Repetier/Printer.cpp
@@ -2337,90 +2337,93 @@ int32_t Distortion::correct(int32_t x, int32_t y, int32_t z) const
 #if JSON_OUTPUT
 void Printer::showJSONStatus(int type) {
     bool firstOccurrence;
+
+    Com::printF(PSTR("{\"status\": \""));
+    if (PrintLine::linesCount == 0) {
+        Com::print('I'); // IDLING
+#if SDSUPPORT
+    } else if (sd.sdactive) {
+        Com::print('P'); // SD PRINTING
+#endif
+    } else {
+        Com::print('B'); // SOMETHING ELSE, BUT SOMETHIG
+    }
+    Com::printF(PSTR("\",\"coords\": {"));
+    Com::printF(PSTR("\"axesHomed\":["));
+    Com::printF(isHomed() ? PSTR("1, 1, 1") : PSTR("0, 0, 0"));
+    Com::printF(PSTR("],\"extr\":["));
+    firstOccurrence = true;
+    for (int i = 0; i < NUM_EXTRUDER; i++) {
+        if (!firstOccurrence) Com::print(',');
+        Com::print(extruder[i].extrudePosition / extruder[i].stepsPerMM);
+        firstOccurrence = false;
+    }
+    Com::printF(PSTR("],\"xyz\":["));
+    Com::print(currentPosition[X_AXIS]); // X
+    Com::print(',');
+    Com::print(currentPosition[Y_AXIS]); // Y
+    Com::print(',');
+    Com::print(currentPosition[Z_AXIS]); // Z
+    Com::printF(PSTR("]},\"currentTool\":"));
+    Com::print(Extruder::current->id);
+    Com::printF(PSTR(",\"params\": {\"atxPower\":"));
+    Com::print(isPowerOn()?'1':'0');
+    Com::printF(PSTR(",\"fanPercent\":"));
+    Com::print(getFanSpeed());
+    Com::printF(PSTR(",\"speedFactor\":"));
+    Com::print(Printer::feedrateMultiply);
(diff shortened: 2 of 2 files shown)


repetier:
That was a bigger patch useful if you have such a display. But I had to modify it quite a bit to fit into the firmware correctly.

1. It uses much FLASH/RAM so it should be optional. You need to set JSON_OUTPUT 1 to have it included.

2. Original version used 810 byte ram plus over 1kb when running some commands. AVR delta users have only around 900 byte free ram. One reason was that you did not utilize flash ram where needed or possible. I have changed that and now it uses only 32 byte ram + 200 byte on stack for some commands.

3. You used new/delete which we do not want to prevent fragmentation of the few bytes left. You even had for M36 file a undeleted new! Changed this to static variables. There was no need for new anyway.

4. Sometimes you reserved memory to escape string to be written. Changed that to write escaped string directly. So no memory is needed and solution is faster.

5. M20 did not work with subdirectories. Subdirectories were not listed in json format. Since the command also allows directly listing a subdirectory I'm not sure how this is supposed to work. I have modified it to only show current directory selected.  But that rises a other question - directories have same output then files, so you can not distinguish them. How is this supposed to work. Would be great if you could check this as I think it is still wrong.

6. I have also added it to config tool in feature section.

7. Next server/host release can handle M20/M36 correctly with your additions in binary mode.

Sorry for the long delay but fixing all our software products to make it work together was quite a bit of work.

toxuin:
Hey, thanks a lot for going through my code! I am by no means experienced in C/C++ and this is in fact my first public commit to an open-source project. I hope you are not mad at me for code quality and I really appreciate your work going through my code fixing it.
I just wanted a feature implemented and there's no better way of getting it done than doing it.
About the subdirectories being listed as the files – if I understand right from the RepRapFirmware code and the Web panel, it is okay for dirs to be listed in the same fashion as the files since every file (or dir) reported in such listing will get consequent "fine info" request which should be responded with {"err":something-but-zero} for directories.

jamesarm97:
Just curious about the comment made on the originating fork. It was said to be tested with the RADDS and RAPMS board yet all the changed code is in the AVR src and not the DUE src folders. Doesn't the RADDS board use the DUE src directory?

repetier:
@toxuin Yes, info for directory return a error message, so that should work. Anyhow please let me know if it is successfully tested.

@jamesarm97 Yes, due has it's own directory. In the firmware I have the code also in the due branch. Infact I develop on AVR and always copy it then to due (except the special due files with differences).

dc42:
The names of directories in the JSON-formatted output should be preceded by a single asterisk character to distinguish them from ordinary files. see http://reprap.org/wiki/Gcodes#M20:_List_SD_card.

toxuin:
There was an error in docs: 'status' field was marked as 'state'. Fixed that. Also, it now correctly outputs the extended JSON output, not only the fields that are "extra" for extended requests (now all responses contain "status", for example).

repetier:
Thanks for the correction. Hope that it now is working perfectly. Also copied the fix to due branch.

dc42:
I recently added a new field to the M408 S0 parameter so that changes to the fan speed are picked up by PanelDue. Here is a sample response to M408 S0 when there is a new response waiting:

SENDING:M408 S0 R0
{"status":"I","heaters":[19.6,20.6],"active":[-273.1,0.0],"standby":[-273.1,0.0],"hstat":[0,2],"pos":[0.00,0.00,150.00],"extr":[0.0],"sfactor":100.00,"efactor":[100.00],"tool":0,"probe":"1","fanPercent":[0.00,0.00],"fanRPM":0,"homed":[1,1,1],"message":"","seq":2,"resp":"Calibrated 6 factors using 13 points, deviation before 0.111 after 0.056\n"}

and when there is no new response waiting:

SENDING:M408 S0 R6
{"status":"I","heaters":[19.6,20.5],"active":[-273.1,0.0],"standby":[-273.1,0.0],"hstat":[0,2],"pos":[0.00,0.00,150.00],"extr":[0.0],"sfactor":100.00,"efactor":[100.00],"tool":0,"probe":"1","fanPercent":[0.00,0.00],"fanRPM":0,"homed":[1,1,1],"message":""}

and here is a typical M408 S1 response:

SENDING:M408 S1
{"status":"I","heaters":[19.6,20.5],"active":[-273.1,0.0],"standby":[-273.1,0.0],"hstat":[0,2],"pos":[0.00,0.00,150.00],"extr":[0.0],"sfactor":100.00,"efactor":[100.00],"tool":0,"probe":"1","fanPercent":[0.00,0.00],"fanRPM":0,"homed":[1,1,1],"message":"","geometry":"delta","myName":"Kossel"}

"probe" is the Z probe reading, 0-1023. if using a digital Z probe such as a switch then I suggest you return 0 for not triggered and 1000 for triggered, which is what RepRapFirmware does.

btw M408 S0 and S1 are used by PanelDue, and S2 and higher by the web server. 

HTH David

repetier:
I guess the new thing is

"fanPercent":[0.00,0.00],"fanRPM":0

but why 2 values for fan and one for rpm? What is the second fan? Sorry  don't know RepRapFirmware behavior so I don't know what 2 fans means.

dc42:
It's because the Duet 0.8.5 supports two PWM-controllable fans, controlled using M106 P0 S### and M106 P1 S### (P0 is the default). I believe that the current web interface (DWC 1.07) will cope with the fanPercent field being either a single value or an array, and the plan is for the M408 S2 and other responses to change to returning an array in future versions of RRF. If Repetier supports only one PWM controllable fan, you can return an array of one element.

repetier:
Sounds good. Also using P to distinguish fans is a good upgradeable solution. Guess at some point I should add a number of fans to firmware/server/host to allow controlling them.

PR #440, opened 2015-11-23
23-11-2015

SigRepetierMod:


repetier:
Why did you close it? Because it was to special and only works with one special printer?

SigRepetierMod:
I just want to follow (keep up to date with) Repetier-firmwork and keep my modifications, but i haven't managed yet.
There are some modifications for one special printer indeed, so i think it's pointless for it to exist here.

repetier:
Yes, I think the same. It is great for the printer that matches, but 99,9% have different printers, so pull requests make no sense.

PR #448, opened 2016-01-06
Wrong pin assignment for Megetronics 3.0

floyd871:
Temperatur Pins for Heatbed and Extruder 2-3 are wrong


repetier:
I can not follow the changes. The request was agains master from work092 so 182 files are changed. Can you say what you think which pins are wrong. The megatronics sometimes think this is wrong and on 10-9-2015 it was even changed to match the board naming. But it difficult since board and firmware naming may differ for some boards.

floyd871:
Sorry for the confusion. I have now created a proper request.

PR #449, opened 2016-01-06
Wrong pin assignment for Megatronics 3.0

floyd871:
Temperatur Pins for Heatbed and Extruder 2-3 are wrong, Correct definition:

#define TEMP_0_PIN 15 // Extruder 1
#define TEMP_2_PIN 13 // Extruder 2
#define TEMP_3_PIN 12 // Extruder 3
#define TEMP_1_PIN 14 // Heated bed

### DIFF
--- a/src/ArduinoAVR/Repetier/pins.h
+++ b/src/ArduinoAVR/Repetier/pins.h
@@ -2201,9 +2201,9 @@ S3(ext)=9
 */
 
 #define TEMP_0_PIN 15 // Extruder 1
-#define TEMP_2_PIN 14 // Extruder 2
-#define TEMP_3_PIN 13 // Extruder 3
-#define TEMP_1_PIN 12 // Heated bed
+#define TEMP_2_PIN 13 // Extruder 2
+#define TEMP_3_PIN 12 // Extruder 3
+#define TEMP_1_PIN 14 // Heated bed
 
 #define THERMOCOUPLE_0_PIN 11
 #define THERMOCOUPLE_1_PIN 10


repetier:
Ok, I see what you mean. You have a different interpretation what should be the heated bed thermistor.
We always assign the last thermistor to the heated bed. To solve a bit the confusion we have changed the comments for next update to:

// Extruder 1 - Thermistor 1
#define TEMP_0_PIN 15
// Extruder 2 - Thermistor 2
#define TEMP_2_PIN 14
// Extruder 3 - Thermistor 3
#define TEMP_3_PIN 13 
// Heated bed - Thermistor 4
#define TEMP_1_PIN 12 

That should at least show how it is used. It is normally a bad idea to change this later as it would invalide all running systems with that board. So I prefer to make it more clear how our names map the thermistors.

Some boards make it a bit easier having a thermistor named for heated bed so it is more clear. Sorry to confuse you with our numbering. Comes from old times where we only had 3 thermistors and outputs and one output was used for fan.

floyd871:
My former firmware i have used was marlin. First there was no Pin defines for the Megatronics 3.0 board. So i have created a pull request from the values provided by the inventor (Reprapworld). In their definement Thermistor 2 was for the heated bed. It makes more sense when the pin for the heated bed is the last one.

repetier:
So some more confusion. But yes, having first all extruder thermistors and then bed makes most sense which is why I did it that way. I guess most users use this intuitively right and only get confused when starting to think about it (or come from marlin).

Will close the request.

PR #457, opened 2016-01-22
Hanno hi patch 2

hannoHI:
Found some peculiarities for distortion correction.


hannoHI:
More than I intended. Will try again

PR #458, opened 2016-01-22
Fix Correction step miscalculation & typo  for distortion correction

hannoHI:


repetier:
I guess you are right with the corrections. Math is not easy :-)

PR #459, opened 2016-01-22
Fix Correction step miscalculation & typo for distortion correction  …

hannoHI:
…(due)


repetier:
Had copied due myself, so closing, Thanks.

PR #460, opened 2016-01-23
Fix command M408 (S2,S3,S4,S5)

Jan-Ruda:
VALID JSON (RFC 4627)


repetier:
Good seen, also I think the newlines are no problem (or at least should be).

Issue #467, opened 2016-02-06
Compile failure with FEATURE_CONTROLLER set


PxT:
I'm unable to compile with the Display controller set to any option > 0.  Configuration below, and compile time errors beneath.  If I set FEATURE_CONTROLLER to 0 then everything compiles fine. I am using the latest github master branch. Arduino IDE v1.6.7 on Mac OS

`{
    "editMode": 2,
    "processor": 0,
    "baudrate": 250000,
    "bluetoothSerial": -1,
    "bluetoothBaudrate": 250000,
    "xStepsPerMM": 80,
    "yStepsPerMM": 80,
    "zStepsPerMM": 80,
    "xInvert": "1",
    "xInvertEnable": 0,
    "eepromMode": 2,
    "yInvert": "0",
    "yInvertEnable": 0,
    "zInvert": "1",
    "zInvertEnable": 0,
    "extruder": [
        {
            "id": 0,
            "heatManager": 1,
            "pidDriveMin": 40,
            "pidDriveMax": 230,
            "pidMax": 255,
            "sensorType": 97,
            "sensorPin": "TEMP_0_PIN",
            "heaterPin": "HEATER_0_PIN",
            "maxFeedrate": 100,
            "startFeedrate": 20,
            "invert": "1",
            "invertEnable": "0",
            "acceleration": 6500,
            "watchPeriod": 1,
            "pidP": 31.36,
            "pidI": 2.18,
            "pidD": 112.9,
            "advanceK": 0,
            "advanceL": 0,
            "waitRetractTemp": 150,
            "waitRetractUnits": 25,
            "waitRetract": 25,
            "stepsPerMM": 92.4,
            "coolerPin": 7,
            "coolerSpeed": 255,
            "selectCommands": "",
            "deselectCommands": "",
            "xOffset": 0,
            "yOffset": 0,
            "zOffset": 0,
            "xOffsetSteps": 0,
            "yOffsetSteps": 0,
            "zOffsetSteps": 0,
            "stepper": {
                "name": "Extruder 0",
                "step": "ORIG_E0_STEP_PIN",
                "dir": "ORIG_E0_DIR_PIN",
                "enable": "ORIG_E0_ENABLE_PIN"
            },
            "advanceBacklashSteps": 0,
            "decoupleTestPeriod": 12,
            "jamPin": -1,
            "jamPullup": "0"
        }
    ],
    "uiLanguage": 0,
    "uiController": 0,
    "xMinEndstop": 0,
[...]

PR #468, opened 2016-02-07
Fix compilation error

PxT:
This resolves the compilation error noted in #467, but I don't know if this is the correct way to fix it or not.


repetier:
That fix does introduce new errors on runtime in error case. Please use my fix.

PxT:
Thanks for the quick update

PR #480, opened 2016-02-23
Add support for Zonestar (Melzi V2.0, LCD 20x4, 5 keys ADC keypad)

osnwt:
There are some printers sold under different names like 'Zonestar P802M',
'Prusa i3 P802M DIY kit', etc, which have LCD 20x4 with 5 keys controller
connected to Melzi V2.0 board via 10 wires cable. Keys are connected to a
single analog input using resitive divider.

Despite GPL license, manufacturers ignore the requirement to share modified
code. So this is a community effort to reimplement the ADC keypad support.

More information on the issue:
    http://forum.repetier.com/discussion/1105/melzi-v2-0-with-lcd2004-and-5-keys

All credits to original modifications and schematics should go to their authors:
 - Hally for the original modified code
 - Mixanoid, Mekanofreak, axelsp for their work on this

This is the attempt to rework the code and possibly merge it into the official
firmware repository.


repetier:
Nice solution for the analog problem. I have copied the relevant parts to the next release. So it will also get a dropdown entry in config tool soon. I changed board id to 24 to get continuous numbering. I also omitted your pictures. They do not really help and I do not know where they come from and it keeps the repository lean.

osnwt:
Sure, we can't keep all schematics in the repository. The number 802 was chosen because there is already an out of sequence number there. Thank you, I hope it will be natively supported some day.

On Melzi there is no more free pins (just one), and I want few more (Z probe, beeper, etc). So I am going to add a shift register mod for the LCD and free 4 more pins. Again, have to add some code to the LCD support. Coming soon.

Thanks for your great software! Sadly, Mac version does not support Repetier-Server (yet?).

szepnorbee:
Hi! Witch GPIO still free on board? Can i use this for Auto bed leveling? Thx

osnwt:
Originally there is just one unused pin on the ext connector (do not remember the number), I don't even know if it is wired to something on the LCD PCB. You can use this one for the sensor or reuse Z-min because it makes no sense if you have auto leveling.

If you need more, you can have 5 spare pins instead of 1. To do so you should add an adapter between Melzi and LCD cable with some simple logic there. Then you may reuse Z-min for Z-probe, unused pin for servo (or beeper), and still have 2 pins for Tx1/Rx1 (wifi web interface or linux host) and 2 pins for SDL/SDA or anything you want. Check these pull requests for details:

This fixes some issues with watchdog on our printers:
https://github.com/repetier/Repetier-Firmware/pull/493

This enables servo support on Melzi:
https://github.com/repetier/Repetier-Firmware/pull/498

This adds a shift register LCD support, so you can use just 2 wires for LCE instead of 6:
https://github.com/repetier/Repetier-Firmware/pull/497

Here is an example of working bed leveling on Melzi with 2 wires for Z-probe, 2-wire LCD connection, and still free 3 pins (including Tx1/Rx1 for soon to be connected linux board):
https://www.youtube.com/watch?v=jJh9SNof-D8

PR #482, opened 2016-02-25
Change debugLevel (toggleEcho)

phd0:
in DUE file


repetier:
Indeed wrong and so long unseen :-)

PR #483, opened 2016-02-27
Zonestar config cleanup

osnwt:
Cleaned up README.txt and sample configuration for Zonestar printer. Also fixed web configurator error which sets the controller number 23 for Zonestar (same as for another controller), while it should be 24.
Please fix the web GUI. Thank you!


repetier:
Config tool is fixed and patch merged.

osnwt:
Excellent, thank you!

PR #484, opened 2016-02-27
add "Debug end stop" on Debug Menu

phd0:
Add a "EndStop" Boolean flag on Debugging menu
Add a dynamic show of state for hardware end-stop
Add com:debug -> serial port of position of end-stop if actives
Add UI_TEXT_DBG_ENDSTOP_ID for text translation

and add some horrible change tabulation and space due "automatic formating text", sorry


repetier:
Nice fix. And gladly your style fixes are as i like them:-)
I have also copied it to the due version and works good.

PR #486, opened 2016-02-28
Add menuLevel 0 (EN and FR only) end-stop state with XYZ coords

phd0:
Show in coords XYZ page menuLevel 0 the stats of the xyzXYZ end-stops
with new special char : %dx ; %dX to %dZ and show only one CHAR (only hardware end-stop)
img_1426

and some retouch on the "go to previous folder" special character with a more long leg


repetier:
Nice idea. Have merged it and internally already added the other languages. Will publish the full version once I have fixed an other bug I'm just hunting.

PR #488, opened 2016-02-29
Expanded distortion correction for cartesian printers

GermanRepRap:
Hi Roland,

Distortion correction is a real cool feature, but it didn’t work on our Cartesian printer. So I made a few changes. Now it works at our printer. (X350)
Splitting segments behaved a little bit strange. It moved up for the first segment and down all the others.
For distorted layers there is now an Extrusion correction. The bed for testing had a distortion of ~0.35mm and the prints are pretty nice.

Best regards,
Andreas

### DIFF
--- a/src/ArduinoAVR/Repetier/Printer.cpp
+++ b/src/ArduinoAVR/Repetier/Printer.cpp
@@ -2203,7 +2203,7 @@ int32_t Distortion::correct(int32_t x, int32_t y, int32_t z) const
 
 int32_t Distortion::correctExtrusion(int32_t x, int32_t y, int32_t z, int32_t e) const
 {
-    //TODO Retracts are multiblied at the moment. This causes strings in layers. (zStart < layers < zEnd)
+    //TODO Retracts are multiplied at the moment. This causes strings in layers. (zStart < layers < zEnd)
     if (!enabled || z > zEnd || z < zStart || Printer::isZProbingActive()){
       return e;
     }
--- a/src/ArduinoAVR/Repetier/Printer.cpp
+++ b/src/ArduinoAVR/Repetier/Printer.cpp
@@ -2181,7 +2181,8 @@ int32_t Distortion::correct(int32_t x, int32_t y, int32_t z) const
       Com::printF(PSTR(" iy= "), fyFloor); Com::printFLN(PSTR(" fy= "), fy);
     }
     if (z > zStart && z > 0)
-        correction_z = (correction_z * (zEnd - z)) / (zEnd - zStart);
+        //All variables are type int. For calculation we need float values
+        correction_z = (correction_z * static_cast<float>(zEnd - z) / (zEnd - zStart));
    /* if(correction_z > 20 || correction_z < -20) {
             Com::printFLN(PSTR("Corr. error too big:"),correction_z);
         Com::printF(PSTR("fxf"),(int)fxFloor);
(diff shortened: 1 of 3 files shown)


repetier:
The distortion correction was corrected yesterday as I found that problem independently, but your solution is a bit better as it prevents the overflow problem I have now.

I also like the extrusion compensation. Will add this soon to main branch.

boelle:
could this be merged to dev branch? i use it a lot but not sure if it will help with a bed that has ~0.5mm disortion

you might say its a lot and it is... its one of those mk42 clones :-/ not happy about it but other than its bad at some spots most of the bed is ok

its mostly at the corners its really bad

repetier:
@boelle Are you already using 1.0? Distortion correction for 0.92 is broken. 1.0 will split moves in 10mm segments and move to interpolated position. In V2 there will be even more corrections once it is working just like for delta printers. 0.5mm is correctable also it would give a wavy bottom surface then and might need more then 1mm to come back to no correction version.

boelle:
i use the dev version

https://www.repetier.com/firmware/dev/index.php

repetier:
Ok that should work. What exactly the problem you experience? Please also post the distortion correction settings (please not all vars)

boelle:
i have spots on my bed where the nozzle is to close to the bed, and there are spots where its to high up

i know 20mm is HIGH, but it was never bothered to lover it

image

repetier:
The max correction is  a safeguard against wrong probing results, so unless you have no bad probings there is no reason to autoreject values anyway. 2mm to compensate 0.5mm - could be a bit squished but ok. After all you want it end fast so the rest is good.

With such a bed you should have 100% correction until first layer height and you should not start with 0.1mm layer, so set start correction better to 0.3 and make a 0.2 or 0.3mm first layer height. That also reduces the errors.

So I hear in general it works but some spots are present. That might also depend on rate of change between points. Since we split moves in 10mm parts and you are 5mm left and right from a peak it will not correct the peak so that would be a close by. So that depends on a bit on luck where lines are split and how much to change over the 10mm. It might even be a problem that it omits one measured value. 160mm in x direction with 22 measurements will just jump over some results for correction. That is why we want in V2 much more updated corrections. The current way V1 handles linear moves make it impossible to do it that way.

boelle:
hehe

i will correct my settings as you are correct. and i will try and be patient and wait for v2 :-D

boelle:
would these settings be better?

Full correction until (FC)
DISTORTION_START_DEGRADE

0,3
[mm]

Reduce correction to zero at (RZ)
DISTORTION_END_HEIGHT

1
[mm]

boelle:
for the fun of it here is how bad it looks

https://imgur.com/a/QSD88

the small X in the middle is the center, the line of the left is where the probe cant reach so should not count from the line to the edge.

inside the circle its almost ok, there are some small variation in the thickness and in close to the center its way to close

but compared to last time i did such a test print the same remains... close to the corners its does not stick very well

repetier:
That is in deed a heavy case,
0.3mm for start is ok, 1mm for end of reduction might be not enough. You correct (1-0.3) = 0.7mm with error of 0.5mm that mean 1+0.5/0.7 = 1,71 => 171% of extrusion in that parts and the filament has to go somewhere.

boelle:
what would help me a lot is to get the corners to the right to stick as well as in the center

of course  the complete left side where the probe cant reach will never be good

repetier:
Would be interesting to see if V2 can do it better when it is working. It will at least make much more updates so follow your map more closely then with the 10mm length.

I'm also thinking of allowing a distortion map over full area in V2 and then only measure what can be measured and use approximaltions at first and allow users to adjust not measured areas manually.

boelle:
that is a good idea since the unprobed part is pretty much the same

so yes a disortion map for the full bed and just probe as much i can then after correct what can be probed manual would be super nice

PR #491, opened 2016-03-01
Zonestar Z axis default step/mm are wrong

szepnorbee:
The correct value is for Zonestar 1600.


repetier:
Ok merged, also this may depend on your microstepping and printer. SO I'm quite sure now someone else has a wrong value:-)

osnwt:
Exactly, and I found that change after complain from my friend :-)

That's why I've created another pull request which **keeps** the 1600 value (as being safer for the printer on low heights) but added a comment about different hardware. Also I've changed the JSON part of config, so it can be reused with Web configurator. It is always a good idea of using Web UI to change Configuration.h, at least, for official code, so the file is (1) consistent, (2) has no changes after up/download if no changes were done (including trailing EOL on EOF).

In the same pull request I disabled Watchdog by default for sample Zonestar config. On my printer and printer of my friend it constantly restarts if no SD card inserted. So safer option is to have it disabled by default.

Here is the pull request: https://github.com/repetier/Repetier-Firmware/pull/493

Also I modded my Zonestar to free up 4 (5 total) GPIO pins to enable extra serial and servo/z-probe pins plus beeper. The pull request is here: https://github.com/repetier/Repetier-Firmware/pull/497

PR #497, opened 2016-03-10
Add support for displays connected using shift register (2/3-wire)

osnwt:
### Patch comments

This patch frees up 3 or 4 wires of 6-wire 4-bit LCD connection by using
well-known shift register mod. That is important to implement extra features
on boards with not enough free GPIO pins such as Melzi.

**The code was tested** on modified Zonestar printer with Melzi board in both
2-wire and 3-wire versions. Both work well, but for 2-wire it was necessary
to use a real AND gate (74LS08) instead of diode/resistor due to long
wires between MCU/SR and LCD.

**The patch changes nothing of original code, it just adds one more type
of display.** An example of use is shown in ui.h for Zonestar controller.
It does not require any change of Web Configurator. Just 3 lines should
be added in user section to enable this version. Of course, they can also
be placed in uiconfig.h if necessary.

### Original commit message

Many popular 3D printer boards have limited number of free GPIO pins.
Say, Melzi board in Zonestar 3D printer has only single unused GPIO.
Few extension pins (including Tx1/Rx1, SDA/SDL) are used for 6-wire
LCD interface (RS, ENABLE, D4-D7).

A clever trick using a shift register allows to free 3-4 of 6 used LCD
pins and connect Bluetooth or Linux board to the extra serial, or some
other extensions such as Servo/Z-probe etc.

This patch adds a shift register connected LCD support driver to the
firmware without using of LiquidCrystal library. It supports both 2 and
3 wire connection. 2-wire option requires an "AND" gate made with a diode
and resistor, or a real AND gate. 3-wire option works without extra
parts.

This code uses de-facto standard wiring of LCD. Example of it
can be seen here:
https://bitbucket.org/fmalpartida/new-liquidcrystal/wiki/schematics

It is compatible with this library:
https://bitbucket.org/fmalpartida/new-liquidcrystal/wiki/Home

### DIFF
--- a/src/ArduinoAVR/Repetier/ui.cpp
+++ b/src/ArduinoAVR/Repetier/ui.cpp
@@ -568,7 +568,164 @@ void initializeLCD()
 }
 // ----------- end direct LCD driver
 #endif
-#if UI_DISPLAY_TYPE < DISPLAY_ARDUINO_LIB
+
+#if UI_DISPLAY_TYPE == DISPLAY_SR
+// Native LCD driver for displays connected using shift register (2-wire or 3-wire)
+//
+// Options:
+//   #define UI_DISPLAY_TYPE        DISPLAY_SR
+//   #define UI_DISPLAY_DATA_PIN    29
+//   #define UI_DISPLAY_CLOCK_PIN   28
+//   #define UI_DISPLAY_ENABLE_PIN  -1 // or undefined for 2-wire, pin number for 3-wire
+//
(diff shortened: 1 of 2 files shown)


roginvs:
I successfully modified my Anet A8 using this patch and one 74LS164N (3-wire version). Now my printer looks like [this](https://pp.vk.me/c633916/v633916041/34964/HYQuGqpcwiI.jpg)

PR #510, opened 2016-03-30
fix typographical error

shohei:
I fixed misspelling in the error message shown while delta initialization fails.


repetier:
Thanks for the fix.

PR #511, opened 2016-04-02
Compiler warning - no return statement for non-void function

andematt:
Compiler warning during build of master. Just changed return type, removed commented out code.

> /tmp/build7aa7be57480c79d9d1f62e5e1de5b63d.tmp/sketch/Commands.cpp: In function 'int digitalPotWrite(int, uint16_t)':
> /tmp/build7aa7be57480c79d9d1f62e5e1de5b63d.tmp/sketch/Commands.cpp:273:1: warning: no return statement in function returning non-void [-Wreturn-type]


repetier:
Will be added in next commit.

PR #513, opened 2016-04-07
Detailed Commenting of Firmware Configuration Settings

EnderWiggin14:
It would be very helpful to have detailed comments regarding what each individual setting or group of settings does for Delta-Arm configurations. The difference between PRINTER_RADIUS, ROD_RADIUS, DELTA_DIAGONAL_CORRECTION, DELTA_RADIUS_CORRECTION, etc. is all very confusing and after hours of searching online I have yet to find anything particularly useful.

It would also be nice to have an explanation of how exactly these parameters affect the 3D-printer. The two auto-calibration programs I have used do not make use of DELTA_DIAGONAL_CORRECTION, which based off of the naming would seem the most helpful. I realize this is the Repetier-Firmware GitHub, but unity between the naming schemes of Reptier-Host and Repetier-Firmware would be extremely helpful.


repetier:
I do not think this was an intentional pull request. It is cross origin and changes 40 files. Not sure what the outcome would be but I'd not trust the result as most commits are already included in master.

Host does not name anything. I guess you mean eeprom names, but these also come from firmware and are made more human readable then config settings.

PR #515, opened 2016-04-09
Work092

RAyWB:
Fixed i2c routines for due , added possibility to chenge clock speed


repetier:
So you figured it out:-) Great.

I have taken your changes with some of my changes into master. Main difference is
i2cSetClockspeed instead of Seti2cClockspeed as it fits existing naming scheme.
I also changed some types but your logic is 100% the same.

PR #518, opened 2016-04-16
Wrong direction ENDSTOP_Z_BACK_ON_HOME move - Update Printer.cpp

madmanisko:
In line:
1291: PrintLine::moveRelativeDistanceInSteps(0, 0, axisStepsPerMM[Z_AXIS] * -ENDSTOP_Z_BACK_ON_HOME * Z_HOME_DIR,0,homingFeedrate[Z_AXIS], true, true);

we need delete "-" before ENDSTOP_Z_BACK_ON_HOME because we have right sign in Z_HOME_DIR (+ or -).

### DIFF
--- a/src/ArduinoAVR/Repetier/Printer.cpp
+++ b/src/ArduinoAVR/Repetier/Printer.cpp
@@ -1288,7 +1288,7 @@ void Printer::homeZAxis() // Delta z homing
 				homingSuccess = true; // Assume success in case there is no back move
 #if defined(ENDSTOP_Z_BACK_ON_HOME)
 				if(ENDSTOP_Z_BACK_ON_HOME > 0) {
-					PrintLine::moveRelativeDistanceInSteps(0, 0, axisStepsPerMM[Z_AXIS] * -ENDSTOP_Z_BACK_ON_HOME * Z_HOME_DIR,0,homingFeedrate[Z_AXIS], true, true);
+					PrintLine::moveRelativeDistanceInSteps(0, 0, axisStepsPerMM[Z_AXIS] * ENDSTOP_Z_BACK_ON_HOME * Z_HOME_DIR,0,homingFeedrate[Z_AXIS], true, true);
 					//Endstops::report();
 					// Check for missing release of any (XYZ) endstop
 					if (Endstops::xMax() || Endstops::yMax() || Endstops::zMax()) {


repetier:
No, that is correct.
All values are positive and we need to move down, so we need a minus to negate the move.

Have you wrongly defined ENDSTOP_Z_BACK_ON_HOME negative?

madmanisko:
@repetier, thats correct - we need minus to negative the move.
My definition ENDSTOP_Z_BACK_ON_HOME is positive (5). 
So at the moment calculation is :  
axisStepsPerMM[Z_AXIS] * -ENDSTOP_Z_BACK_ON_HOME * Z_HOME_DIR
look, we have to possibility to calculate this formula (HOME to minimum = Z_HOME_DIR is 1 or HOME to maximum = Z_HOME_DIR is -1)

First version (HOME to minimum) we need to calculate positive steps (move up):

axisStepsPerMM[Z_AXIS] is positive
ENDSTOP_Z_BACK_ON_HOME is positive so -ENDSTOP_Z_BACK_ON_HOME is negative
Z_HOME_DIR is positive (1)
this formula is (+)*(-)*(+) = (-) result is negative not positive

Second version  (HOME to maximum) we need to calculate negative steps (move down):

axisStepsPerMM[Z_AXIS] is positive
ENDSTOP_Z_BACK_ON_HOME is positive so -ENDSTOP_Z_BACK_ON_HOME is negative
Z_HOME_DIR is negative (-1)
this formula is (+)*(-)*(-) = (+) result is positive not negative.

Am I right, or i something not understand with this variable.
After delete this minus my problem with "pause printing" when i home axis G28 (delta printer) gone.

repetier:
Deltas have Z_HOME_DIR 1 so it is positive. We home to z max after all. So with your logic we have 
(+)(-)(+) = (-) like you described in home to minimum, also the description is for home to maximum. At the top we have maximum z!

madmanisko:
You have right (i'm wrong) ;-). But it doesn't matter if i set Z_HOME_DIR to 1 or -1. My delta always home to max. Both configuration makes move head to max position. I mean i change in config file only Z_HOME_DIR nothing else. I must check this in the future.

I think that you should close this pull.

repetier:
You are also right. It is the only line in delta homing using the direction at all. Should be removed in future as we know homing direction anyway.

PR #542, opened 2016-06-26
Fix parsing of extended gcode arguments and remove extra G0 laser-off from G2/G3

dewhisna:
These changes fix the variables used for parsing the extended gcode arguments C, H, A, B, K, L, and O, instead of incorrectly parsing into 'D'.   And removes the useless laser-off guard for G0 commands in the G2/G3 processing case, as the 'if' is always false.

Note: this should be merged to the 'development' and 'work092' branches as well, as they both have the same issue.

### DIFF
--- a/src/ArduinoAVR/Repetier/Commands.cpp
+++ b/src/ArduinoAVR/Repetier/Commands.cpp
@@ -908,19 +908,7 @@ void Commands::processGCode(GCode *com) {
 #if ARC_SUPPORT
         case 2: // CW Arc
         case 3: // CCW Arc MOTION_MODE_CW_ARC: case MOTION_MODE_CCW_ARC:
-#if defined(SUPPORT_LASER) && SUPPORT_LASER
-            {
-                // disable laser for G0 moves
-                bool laserOn = LaserDriver::laserOn;
-                if(com->G == 0 && Printer::mode == PRINTER_MODE_LASER) {
-                    LaserDriver::laserOn = false;
-                }
-#endif // defined
-                processArc(com);
-#if defined(SUPPORT_LASER) && SUPPORT_LASER
-                LaserDriver::laserOn = laserOn;
-            }
-#endif // defined
+            processArc(com);
             break;
 #endif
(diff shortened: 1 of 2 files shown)


repetier:
Ok, have patched the development version also for due. We don't patch against master to not obscure modification history.

dewhisna:
Ah, sorry.  I wasn't sure what your development flow was.  I had it on 'master' because I was working on merging the latest 0.92 release with my repo branch for the BoXZY printer customization.  Plus, I figured you could simply change which branch you merged it into if you wanted it on a different one.

Next time I'll make the change on 'development' and cherry-pick it into my copy instead.  Also, is the 'work092' branch still active or has it been displaced by work in 'development'?  I ask because it looked fairly current and needed this set of patches as well.  Thanks!

repetier:
No problem, was a small patch so I redid it by hand on development.

work092 is expired and not used any more. We now switched to development/master as pattern. All changes in development and from time to time if I fill it validates for a new version I push all to master. That will also make more sense as next release will get 1.0.0 version number.

Maybe it is possible to move merges to development when made to master, but I'm not very familiar with all git tweaks and I also don't trust that as push requests are modifications against a different code base. So expect that to be a problem.

dewhisna:
Actually, you'd be surprised at how good git is with that.  Each commit is essentially a delta (think 'patch' file), so even if it were done with a very different code base, it would still merge correctly.  And even more trivial in this case, as my code repository was actually even with the official repository on all branches.  The only thing that would have been tricky was that there was a minor merge conflict on the 'development' branch from other code change work that was in close proximity to the change I had, so it would have stopped with a conflict, which you'd resolve by hand, and then commit after fixing that one line.

But your redoing it is OK.  Now, it will make a merge conflict for me and my repository the next time I update, since we both made the same change on separate commits and won't have recorded the merge from my change (especially since I already merged my commits with my other work branch), but it will be easy enough to resolve since it was a small change.

Now that I understand your development/master setup, I'll do future pull requests on 'development'.  Thanks!

PR #554, opened 2016-07-23
Allow M117 commands in SD prints

gonzalezjj:
This patch allows using M117 commands in gcode files printed directly from an SD card. I'm currently using this to display the remaining time as calculated by my gcodetimer tool (https://github.com/gonzalezjj/gcodetimer) and thought it could be something others might be interested in.


repetier:
Will be included in next development update.

gonzalezjj:
I forgot to add the same changes for the DUE codebase. See pull request 555

repetier:
No problem I already have it in dev version for both. 

BTW: We do not take pull requests agains master - all changes go to develop first and from time to time we push everything over to more stable master branch.

PR #555, opened 2016-07-25
M117 in SD mode (updated for AVR+DUE)

gonzalezjj:
Update to pull request 554. Added changes for both AVR and DUE boards

This patch allows using M117 commands in gcode files printed directly from an SD card. I'm currently using this to display the remaining time as calculated by my gcodetimer tool (https://github.com/gonzalezjj/gcodetimer) and thought it could be something others might be interested in.


repetier:
Already included:-)

PR #557, opened 2016-08-02
High temp thermistor dev

JeanSebastienCarrier-Dyze:
These modifications allow the use of high temperature thermistors in order to handle hot ends temperature up to 500 degrees Celsius.

Because thermistors able to read higher temperature tend to have a hard time reading lower values, it was necessary to allow a "preheating time", else a temperature error would be thrown.

I added a new table (NUMTEMPS_15 27) with the value we're using for our thermistor.

It is true that the best case scenario would be to use the proper amplification circuit with a platinum or thermocouple sensor, but I prefer a drop-in remplacement with minimum hardware add-on and changes.

All sensors types have their advantages and disadvantages. Since it is possible to do a little trick on the firmware to go around the weakness of the high temperature thermistor ( low resolution below 40°C ) , it makes this solution more than interesting;

Better resolution than an amplified PT100 from 100°C to 400°C (Ultimaker circuit Op-Amp)
Better resolution than K type thermocouple from 140°C to 325°C (using AD595)
No add-on circuit required
Faster response than PT100
Very stable
I think these sensors have been underrated due to their limitation at low temperature and I think it is very well suited for 3D printer application.


rcarlyle:
This is interesting. Do you have a datasheet you can share on the thermistor you're using?

Some people (foolishly) run hot ends that can achieve unpleasantly high temperatures in 30 seconds, so I would recommend being cautious about default values and such...

kyrreaa:
My MightyMouse hotend (5 grams) reach 240c from 20c in 9 seconds...
I use a thermocouple though, and set the readout intervall shorter to avoid unstable pid problems.

K

alfredanil:
Will PT1000 work without any add-on circuit? for sure it can offer better resoulution than PT100.

woodencase01:
@rcarlyle The thermistor used for this hotend has no specsheet available, but we have RT tables. AdSem used to make such thermistor, but it seems like the website is down.
You are right about the values, maybe a maximum value should be suggested in the comments. 15 seconds should be plenty for any applications.

@kyrreaa Very interesting, 5 grams is very little. The DyzEnd is quite small, but it is far from 5 grams. This must probably not include the fan, sensor cables and heater, right?

@alfredanil The PT1000 is not bad, reaching a resolution of ~1.8°C at 20°C down to ~2.4°C at 250°C. It is okay, but not ideal for 3D printing because there is quite a big margin of error. However, you are totally right, a PT100 is totally non-viable solution, with a resolution of 14°C...

This high temperature thermistor allows to reach a resolution of 0.16°C at 200°C, which is at printing temperature. Stability is at its best.
The downside is the poor resolution at room temperature, which can be easily fixed with a clever firmware trick. No add-on circuit required. It seems like a very interesting solution when taking performance and price into consideration.

Hope it helped!

repetier:
Made some adjustments especially for due and published it also in online config tool.

kyrreaa:
woodencase01:
It includes heater but not ptf fitting.

K
3. aug. 2016 5.27 p.m. skrev "woodencase01" <notifications@github.com>:

alfredanil:
temperature-resolution-comparison-for-sensors-used-in-3d-printers
Explains everything

alfredanil:
960K NTC thermistor from Vishay could be another possibility
[high_temper_thermistor_ntcv101e4964hmb0.pdf](https://github.com/repetier/Repetier-Firmware/files/401808/high_temper_thermistor_ntcv101e4964hmb0.pdf)

woodencase01:
@alfredanil Many thanks for pointing the graph and suggesting the sensor.

The NTCV101E4964HMB0 is interesting, however the resolution isn't as good as our current thermistor. This sensor wouldn't actually require a firmware fix to work correctly, but the resolution would drop starting at 150°C. It is a little bit less interesting for 3D printing. Also, the price is much higher than other thermistors.

rcarlyle:
The problem you'll run into with the "high resolution" window for a thermistor is that's also the least repeatable range due to manufacturing variability and curve fitting errors. The steeper the resistance/temp function (which is where you get the resolution), the less accurate the calibration will be from thermistor to thermistor in that temp range. Which may or may not be an issue for you, I don't know.

I'm not real familiar with the ADC handling code -- what's the effective resolution of the ADC reads here (including oversampling etc)? 10bit? Or better than that?

Honestly, if you want to measure a span from 20C to 450C with usable resolution, I would argue that 10bit analog input hardware is simply inadequate, anything you do to make it work is a kludge, and you should switch to an SPI thermocouple amp like the MAX6675 or MAX31855. But that's my opinion. Reasonable people can differ on that.

alfredanil:
at 350°C, DYZE 500 has a resolution of ~0.75°C whereas pt1000 gives ~3°C - how much does this matter? compare it to the cost of DYZE 500 in EU is about 25.5 USD (10USD shipping included) where as pt1000 thin film type on ebay costs ~2 USD incl. shipping. Please justify cost to resolution at 270°C-350°C range.

kyrreaa:
This is perfect example of why I like the k-type thermocouple. With the
compensated amp I get 5mv/c all the way which on 3.3v ref maxes out at
660c. If using low ref i divide and get 2.5mv/c making 720c max for 1.8v
ref.

K
4. aug. 2016 13.48 skrev "Anil Kumar" <notifications@github.com>:

PR #564, opened 2016-08-14
fix linker error for Arduino with LTO compilation

corecode:
Because stepperWait is only referenced by assembly code, the compiler
thinks that it can optimize it away.  We need to mark the variable as
used, so it sticks around.


repetier:
Thanks.

PR #567, opened 2016-08-22
Adding support for the SAV MK1

boelle:
Added at the end of the pins file as board number 888


repetier:
I added the board to dev version and in online config tool thanks.

BTW: We only take commits against dev version, therefor I copied it by hand to it.

boelle:
oki

wil try and remember that.

there might come a PR that adds support for an Oled board for the same controller

repetier:
No problem. Just give it the next free controller number. I assigned board id 89 btw.

boelle:
oki

its not a controller but a display :-D

PR #577, opened 2016-09-14
second display initialization should not execute MCP23017 init code when

jsb1:
A configured PCF8574 gets MCP23017 init code while initializing display. PCF8574 has no control registers. It simply writes those bytes out to port. This can cause damages.
I could not test a real PCF8574-Display for now, but the current code is dangerous.


repetier:
I have added the following test for next update instead
#if UI_DISPLAY_TYPE == DISPLAY_I2C && defined(UI_DISPLAY_I2C_CHIPTYPE) && UI_DISPLAY_I2C_CHIPTYPE==1

just to be sure we are compiling for I2C at all.

PR #612, opened 2016-12-12
Disable beeper even if defined by board once disabled by FEATURE_BEEP…

sanchosk:
Disable beeper even if defined by board once disabled by FEATURE_BEEPER set to 0.
This will get rid of the annoying quiet speaker sounds :)


repetier:
Included in next update.

PR #622, opened 2017-01-13
fixed bug - doesn't compile when FEATURE_Z_PROBE is off

andyhomes:
Hello
I've found a bug while playing with different config parameters.
Couldn't compile sources when FEATURE_Z_PROBE set to 0
Fixed in both platforms.


repetier:
Thanks for the idea. Added it.

PR #625, opened 2017-01-14
fixed bug - software leveling could not be disabled

andyhomes:
Hello
I just noticed that it's not possible to disable SOFTWARE_LEVELING feature on my delta, so fixed it in both platforms. Now one can switch it off.


repetier:
Good point seen here.

PR #627, opened 2017-01-16
Measure z-probe height feature

andyhomes:
Hello

There is an idea to add "Measure Z-probe height" menu item right after to "Set Z=0" in delta calibration menu.
So that, any time when one is sure printer has correct height set, he can run this procedure and z-probe height is measured and stored into EEPROM. In general this would be used once per delta setup - set Z=0 ("paper test" or any other), then measuring probe height.

The menu item appears only when FEATURE_Z_PROBE is enabled.

The procedure can be also run with "G30 S" command (presence of "S" attribute discards any other behaviour of the command).

Note that Printer::runZProbe method implementation has been changed. 
I never had success with original implementation. It always returned weird numbers, I had no clue on how to interpret them. 
Provided changes returns proper values of z-probing on my delta.
Maybe I do not understand something, because original implementation stays the same for years, but then it's something hard to get. I always used mine since I started using Repetier FW 2+ years ago.


repetier:
So if I understand the idea correctly, you set z=0 and then go up to untrigger z probe and then call the function measuring a z with probe and use difference to set a new z probe height. Good idea whcih gives me some more ideas. 
1. Should not only be for deltas, makes also for cartesian sense.
2. G30 Zx says I'm now exactly at x mm height make sure z probe height is set for this. I like a method with a metal block instead of paper so for me it is always 5.97mm:-)
3. I think we should extend the menu entry to allow setting that height as well. Means a wizard screen with current z as start point allowing change in 0.01 increments to set that value. That is now very easy, did it for felix printer with a wizard variable as tmp var. We now have placeholder to show that value. I think that makes the function even more flexible.

What do you think?

andyhomes:
Yes, the idea caught right.
I do following steps:
a) heat printer's bed and hot-end 
b) calibrate z=0 on a test tool (anything you like, paper, business-card, metal block etc) the tool's thickness doesn't really matter. (yes, the printers height is not correct at the moment, but will be set properly after zp height measurement, see step 'd')
c) Run "Meas. Z-probe height" either from the menu or with "G30 S" (according changes in this pull-request). Let z-probe hit the same tool's surface that was used to set z=0. Now the height is measured and gets stored into EEPROM.
d) Run "G30" and let z-probe hit bed's surface... now printer's height is correct. 

So, right after clicking on "Set z=0" menu, I click next - "Meas. Z-probe height" and it does all the necessary movements - goes up, deploys z-probe (both in z-probe start script), then measures and undeploys. G30 is performed before each print (it's in the print's G-code file)

> Should not only be for deltas, makes also for cartesian sense.

The feature I proposed relies only on FEATURE_Z_PROBE definition, so that should work on any machine having z-probe.

> G30 Zx says I'm now exactly at x mm height make sure z probe height is set for this. I like a method with a metal block instead of paper so for me it is always 5.97mm:-)

I'm afraid I do not understand you. I do not see Z attributes  of G30 command. Checked in different branches, there is no one. Could you explain what do you mean?
I definitely do not understand your method of working with z-probe.

> I think we should extend the menu entry to allow setting that height as well. Means a wizard screen with current z as start point allowing change in 0.01 increments to set that value. That is now very easy, did it for felix printer with a wizard variable as tmp var. We now have placeholder to show that value. I think that makes the function even more flexible.

It would be nice to have, but not really necessary, because one can always tweak the value
M206 T3 P808  X17.45  ; Z-probe height [mm]

repetier:
G30 Zx is not implemented. I meant it as extra parameter to your S in case you use the metal block, see my description for z probing I wrote yesterday:

Calibrating Z-Probe

One thing that is normally wrong at start is calibration of z probe. So that should be done before you start using it to calibrate everything.
First check is always if signals are correct. So send M119 and see that z probe shows a “L” for low = not triggered. Now trigger it by hand while sending M119 again. Now probe value should show a “H” for high = triggered. If it is the other way around you need to change Z_PROBE_ON_HIGH. If nothing changes you have either the wrong pin used or configured, or pull-up must be different. Fix it and continue.

The most important variable is Z_PROBE_HEIGHT which you normally adjust in EEPROM until it has the correct height. It defines the distance between the nozzle tip and bed
at the time the z probe switches to high. For nearly all types this will be a positive value. Once z probes that use force feedback sensors or have a switch triggered by pressing the extruder have 0 or a small negative value. There are several ways to determine the correct height, but the one I now show is the most accurate I know so it should be preferred. You need a flat metal block where you know the height of the block. The blocks height is our reference height (ref_height). First we go close to the
bed at a distance that the z-probe does not trigger and send a

G30 P0
The returned z height is memorized as z0 for later. Now we heat up nozzle to working temperature and remove all ooze. We leave it at temperature so we measure distance to nozzle tip and not to the end of hardened plastic. Now we send

M114
And note the z height as Zstart. Next we adjust z until the metal block fits exactly below the nozzle. You will see that when you have a gap you can easily rotate the block. If you are too low, it will give a “clong” sound when you press bed down. This makes it much more exact then the typically used sheet of paper. Once you think you found exact level, send M114 again and note it a Zend.

So what we now have is starting height as firmware would assume it from first probe Z0 using our initial Z_PROBE_HEIGHT (ZHold). We also know that we needed to go up Znew-Zold to reach ref_height. From this we can compute the exact real value required from Z_PROBE_HEIGHT (ZHnew):

ZHnew = ZHold + Z0 + Znew - Zold - ref_height
Enter this value in eeprom and your measurements will be as precise as you can get it.

Hints: This works fine if you have always the same bed and bed coating. Inductive sensors measure distance to the metal bed part, so it would be possible to correct different bed coatings that you can but on it. The current thickness is stored in Z_PROBE_Z_OFFSET (bed coating in eeprom) and should be 0 for most printers.


As you see it is exactly the same measurement, but you call not at z=0 but at a different height which you know it is at and set it in display. Actually it would void the need for your z=0 function as you would in your case select z = 0, while for my block form the example I select 5.97mm. Rest would be the same. Benefit is that is easier to understand from menu for users not reading manuals (at least until I write it).

I think I will pull your request as soon as my current test/fix is finished and add it. Should not be too much work, as I use your routines with a additional z parameter. I hope you understand when you see.

andyhomes:
Thank you for the explanation.

It looks quite complicated, but I will play with this approach and maybe find some mistakes in my conclusions.
The main concern is about implementation of Printer::runZProbe() I will recheck it. Please do it also before you pull it, I didn't validate in other places of use except G30 command, so maybe some changes needed.

At the moment I see possible issues with my scenario of setting up z-probe height - it's only good for "physically-touching" probes and not good for IR/inductive sensors, where bed coating and the measure "tool" materials are important.

Alternative scenario could be following:
1. perform all the necessary checks etc
2. heat bed and hot-end
3. Set nozzle on tool surface (paper or metal block)
4. set z=0 (optional, this will help to do ajustments at next step)
5. remove the tool and lower nozzle position down for known distance (paper - 0.1 mm, your metal block - 5.97 mm)
6. set z=0 again
7. select "Meas. Z-probe height" - this will take real probe.

repetier:
I will surely check your runZProbe version and see if it should give same result and if not which one is wrong. I know it is a bit complicated as repeated moves do not go up to original start and that can make easy any of our computations wrong.

andyhomes:
> First we go close to the
bed at a distance that the z-probe does not trigger and send a
G30 P0

This is not universal and dangerous recommendation. This is only for "always ready" probes, but not for ones that are deployed with Z_PROBE_START_SCRIPT, since P0 ignores start and finished script.
So that nozzle smashes bed.

What is intention of P0 in this case? Not to copyLastCmd in updateCurrentPosition?

Do I get wrong probe height measurement by your method using G30 (that's actually G30 P3) at this step?

repetier:
You are right. Have no probes that depley probe so I tend to forget it.  Intention is that it measures where you are and not with an offset. But if that is important it will fail that way.

repetier:
Ok, have rechecked your runZProbe and I think it is completely wrong:-(
It does not depend at all on Z_PROBE_HEIGHT at all. You can set that any value it it is simply ignored. If you homed before it will give something around Z_PROBE_HEIGHT I guess.

What I do is in lastCorrection I store starting height in steps. Then go down and count steps x times. Everytime I add lastCorrection-setpPosition which is the height measured from start to trigger. As we add it x times we divide by x and add Z_PROBE_HEIGHT as this value was neglected. Otherwise it looks good and I will now start merging.

andyhomes:
Yes. My implementation of runZProbe returns value that reflects nozzle position at z axis when probe triggers. It's near  Z_PROBE_HEIGHT.

I would suggest to introduce another method, say getZError() or something, that takes probe and subtracts Z_PROBE_HEIGHT, this way you have proper error value and the same time it easily lets use runZProbe to measure Z_PROBE_HEIGHT

With your implementation of runZProbe I never could get proper values, and I still didn't get why.

repetier:
You seem to do things different. Just updated blindly to get some compile errors solved and also changed the measurement method. Currently I can not understand why it should work and hope my solutions works. Still need to test this tomorrow and make fixes if necessary.
Maybe the different idea what runZProbe should return is the reason it works for you and with original runZProbe not. One thing Do not understand is that you wanted to run this at z=0, but that is not possible for my delta for example. z probe height is > 0 and at z=0 it is already triggering for some millimeters, so it needs to go first up to probe, which is what I also added and needs testing as this can conflict easily with autoleveling and distortion correction. Hope to finish this tomorrow with some tests.

andyhomes:
> Maybe the different idea what runZProbe should return is the reason it works for you and with original runZProbe not. One thing Do not understand is that you wanted to run this at z=0, but that is not possible for my delta for example. z probe height is > 0 and at z=0 it is already triggering for some millimeters, so it needs to go first up to probe, which is what I also added and needs testing as this can conflict easily with autoleveling and distortion correction.

Everything to go to safe position before z-probing is in Z_PROBE_START_SCRIPT in my config. That's why I can run probing at any Z position of the nozzle.
`#define Z_PROBE_START_SCRIPT "G1 Z100 F9000\nM400\nM340 P0 S2050\nG1 Z30"
#define Z_PROBE_FINISHED_SCRIPT "G1 Z75 F9000\nM400\nM340 P0 S700\nG1 Z150 F9000\nM400\nM340 P0 S0"
`

andyhomes:
Forgot to tell that autoleveling and distortion correction are disabled in my config.

repetier:
I see that you had some special conditions and a enable command that can influence results. Have now updated the leveling and found some more issues that are now hopefully fixed. At least within the limited precision of my test printer it seems to work well.

andyhomes:
I have just tried with last updates...
G28 and then G30
This caused hot-end move to over extreme position on -Y axis.
I think the problem is in this piece (of Printer::startProbing() method):
`    if(currentPosition[Z_AXIS] > maxStartHeight) {`
`        moveTo(IGNORE_COORDINATE, IGNORE_COORDINATE, maxStartHeight, IGNORE_COORDINATE, homingFeedrate[Z_AXIS]);`
`    }`

`moveToReal` must be instead of `moveTo`, right?

PR #646, opened 2017-03-14
Add support for RRW Graphical LCD

brupje:


repetier:
Ok, have tried to integrate this in a compatible way. Display id is 27 since 25 was already used and we can not change afterwards, that would make old configs invalid.

Only thing not sure is the sd card handling. Since you have no hardware SS pin on the board it now seems like a mix of several attempts and you globally changed spi handling which would most likely cause problems on other boards.  I think the solution active now is hardware spi in sdfat but on HAL this is software spi so it can handle the non standard SS. Please test if it works that way. 

The new display is now listed in config tool.

brupje:
>Ok, have tried to integrate this in a compatible way. Display id is 27 since 25 was already used and we can not change afterwards, that would make old configs invalid.

Fair, I think you want to reflect this in configuration.h as well though. 25 and 26 are missing now.

> Only thing not sure is the sd card handling. Since you have no hardware SS pin on the board it now seems like a mix of several attempts and you globally changed spi handling which would most likely cause problems on other boards. I think the solution active now is hardware spi in sdfat but on HAL this is software spi so it can handle the non standard SS. 

Root problem is that u8glib v1 does not support hardware SPI on DUE boards, therefor you need to fall back to software SPI when using the graphical LCD. v2 should work as it uses the normal Arduino SPI library, but migrating is not easy. I attempted to make this reflect in HAL.h

>Please test if it works that way.
It does not work unfortunately. The SD is still conflicting with the temperature (MAX31855). Ejecting the SD card makes temperature read again. Somehow the LCD is fine though, probably because it does not need to read from that device. 
SDSS is still 59, 

I am not sure why the HAL implementation is so different from the Arduino SPI library, which implements it nicely I think. Why does HAL touch the SDSS pin anyway in software SPI? It should only touch it when SD card it addressed in SDFat. Why not make it accept a CS pin when call spiBegin(transaction), and work with that until  spiEnd(transaction).

repetier:
Originally spi was only used for sd card which leads to several problems with hardware spi. In addition on due the processor uses hardware signaling on CS pins and only 4 special pin numbers are available as slave select pin. SO in the long run we surely need to rework the spi handling in firmware.

PR #647, opened 2017-03-16
Double Endstops::Update

robbycandra:
Typo


repetier:
For next update I have fixed readme type.

```
    Endstops::update();
    Endstops::update(); // need to call twice for full update!

```
is correct. To help a bit against crosstalk you need to trigger signal twice to have an effect!

PR #649, opened 2017-03-18
Remove double defined drive system. This already defined above at lin…

robbycandra:
Remove double defined drive system. This already defined above at line 83.


repetier:
Good catch. Fixed for next update.

PR #650, opened 2017-03-19
Configuration_for_MEGA_and_RAMPS.h

efrain-atienza:
There're many cases where Melzi board is failing, so it's very common to replace it with another option. This is the configuration for one of the most used options: MEGA & RAMPS.

The configuration is prepared for these options:

- RAMPS 1.4
- Arduino MEGA 2560 R3
- Drivers A4988
- LCD12864 3D Printer Controller Board Kit


repetier:
I also add it in dev tree for next update.

PR #654, opened 2017-03-22
Fix Printline Dir

robbycandra:
@repetier, please merge this :D

### DIFF
--- a/src/ArduinoAVR/Repetier/motion.cpp
+++ b/src/ArduinoAVR/Repetier/motion.cpp
@@ -1809,7 +1809,7 @@ inline uint16_t PrintLine::calculateNonlinearSubSegments(uint8_t softEndstop)
 #endif
                     d->deltaSteps[i] = static_cast<uint16_t>(delta);
                 }
-                else
+                else if (delta < 0)
                 {
                     d->setMoveOfAxis(i);
 #ifdef DEBUG_DELTA_OVERFLOW


repetier:
This change omits the case delta = 0! So d->deltaSteps is undefined and can cause any error. Also it slows down computation with extra test. For 0 steps direction does not matter and current solution still stores 0 steps instead of undefined.

robbycandra:
Hi @repetier . Here is the case:

X axis touch the x-min endstop
At this state, I want to move Y axis only.

Even Y axis isn't touch any endstop, it cannot move, because X axis show negatif direction and xmin Endstop active.

I only can move X axis to positif, cannot move X to negatif and cannot move Y axis in both direction.
Therefore, X axis which is not moving, cannot set dir to negatif direction. Cause it will stop other axis.

Regards

repetier:
While I see what you mean, your solution just makes the same error now happen at x max endstop!
In fact you never should touch a endstop during print. If you do so you are lost anyway as move stopes at the moment the endstop is registered, so from that point all positions are wrong. So real solution would be preventing to to go to endstop position. For homing we have therefore even the possibility to move after triggering - one main purpose of this is to untrigger the endstop.

Do you still think that is needed? Please describe why and where so I understand it. We also have ALWAYS_CHECK_ENDSTOPS so you can ignore them during normal operation if that would lead to problems (or if crosstalk likes to trigger it).

robbycandra:
Offcourse it still needed. 

--- "While I see what you mean, your solution just makes the same error now happen at x max endstop!"

It certainly did not happen

Your version -- if delta is zero, X point to negative direction. 
because X point to negative and it touch to min X endstop -> it stop movement on other axis.

Using my version, X did not point to any direction. 
so even therefore even X touch min X endstop -> I still can move Y axis.

Why it is needed ?
Because I need to move joint separately. 
All robotic arm have to type of movement, 
- base on joint angle
- base on cartesian position.
It is terible if after x touch endstop, I cannot move other joint.

And
1. I will not make any error in delta configuration
2. I will reduce speed.

If you still need more explanation I will record a video, to compare both code in a real life robot.

Regards.

PR #656, opened 2017-04-01
Master

gozzzart:


repetier:
Seems like these changes are already in dev tree.

PR #660, opened 2017-04-17
Fix Watchdog reset loop

fullTalgoRythm:
Stop watchdog implementation in HAL.h is missing from Due side. This can cause a reset loop when hw setup takes a long time such as with 12864LCD.

### DIFF
--- a/src/ArduinoDUE/Repetier/HAL.h
+++ b/src/ArduinoDUE/Repetier/HAL.h
@@ -840,7 +840,9 @@ class HAL
       WDT->WDT_MR = WDT_MR_WDRSTEN | WATCHDOG_INTERVAL | 0x0fff0000; //(WATCHDOG_INTERVAL << 16);
       WDT->WDT_CR = 0xA5000001;
     };
-    inline static void stopWatchdog() {}
+    inline static void stopWatchdog() {
+      WDT_Disable(WDT);
+    }
     inline static void pingWatchdog() {
 #if FEATURE_WATCHDOG
       wdPinged = true;


repetier:
Ok have merged it and hope it works. According to ARM documentation you can not disable a watchdog once enabled. SO this might cause watchdog to never enable. Will test it and if it does not work remove commit.

repetier:
Ok, tests showed that your trick does not work. It disables watchdog forever just as I feared. So you have to tune lcd setup to be faster then the 4 seconds, which should be no problem if you do not add too much delay.

fullTalgoRythm:
Thanks for testing it. I meant to double check this on my side as I thought something like this might happen. I'm afraid I forgot, I have another solution as I looked at 2 different methods at the same time. Maybe worth changing this to a comment instead so anyone who looks will realise it's left out on purpose.

fullTalgoRythm:
Looks like the issue I experience is actually caused by trying to restart the watchdog timer when it is above WDD i.e. after a reset. You already have a workaround which is setting WDD to the maximum value.

If you are interested then the correct solution appears to be
```
      WDT->WDT_CR = 0xA5000001; //reset clock before updating WDD
      delayMicroseconds(92); //must wait a minimum of 3 slow clocks before updating WDT_MR after writing WDT_CR
      WDT->WDT_MR = WDT_MR_WDRSTEN | WATCHDOG_INTERVAL | (WATCHDOG_INTERVAL << 16);
      WDT->WDT_CR = 0xA5000001;
```

PR #662, opened 2017-04-20
Added support for SMART RAMPS (EEPROM) + AZSMZ 12864 LCD + BT 

cxandy:
http://reprap.org/wiki/File:SR_RepetierDue_20170418.zip


repetier:
Master is still 0.92.9. Not going to move 1.0 to it yet and this would merge nearly everything.

PR #663, opened 2017-04-22
Development

Braccoz:
fixes for dual x axis stuff.

despite these, there are still issues with dual x when LAZY_DUAL_X_AXIS is disabled, so additional fixes will be needed


repetier:
Ok, see you have bound homing to first extruder direction needed. Good solution. Code is now merged and online.

What issues are left?

Braccoz:
if you print without LAZY_DUAL_X_AXIS somehow the offsets get doubled at some point and it results in bad crashes between the carriages

PR #666, opened 2017-04-22
Fix DUAL X with LAZY_DUAL_X_AXIS disabled part 2

Braccoz:
as remaining issues with dual_x are related to re-selection of current extruder it's better to just skip executing the code. 
To not break any other setup that may still need (altough odd) the old behaviour it's just limited to DUAL_X_AXIS == 1

remade pull request as i forgot DUE files


repetier:
Yes, agree that it might be unnecessary to constrain, but with all these special cases I may easily overlook a case, so better leave it there.

PR #667, opened 2017-04-24
Changed the fix for DUAL_X without LAZY_DUAL_X_AXIS.

Braccoz:
On a second thought, it is better for consistency to not skip the whole rest of the function for just 1 setup and not the rest


repetier:
Ok might be better until I know why I run through all this.

PR #668, opened 2017-04-25
Dual Endstops on X/Y

ikidd:
Not a final PR, not fully working.

Copied MULTI_ZENDSTOP_HOMING for X and Y axis.

Currently NOT using X2_MIN and Y2_MIN in EXTENDED_ENDSTOPS to reduce variables for testing purposes.  Moved X_MAX and Y_MAX to free up flags.  I figured since ENDSTOP_Z2_MINMAX_ID was in the first byte, there might be a reason for that and avoided the second byte until it was working.

Code works fine in Y, not in X.  Y gracefully squares the axis, X grinds both motors.  Confirmed it was a coding issue by swapping all the X/X2 and Y/Y2 stepper and endstop pins in config, then the X axis works fine and Y drives both motors all the time. 

Also confirmed that even a clean copy of Dev uploaded with a working config from Master acts as if there are software min endstops until homed, on ALL axis.


repetier:
Something seems wrong with your pull request. If I look into files changed there are 0 changes.

There is no reason ENDSTOP_Z2_MINMAX_ID is in byte 1 except that it had space at the time I needed it. Maybe you should first start with a simple modification. Extend endstop handling to handle x2/y2 min/max in update and report. Then M119 allows you to trigger all endstops manually and see if they get signaled correctly. If they do continue with your dual endstop for homing code. Since you have no changes I in summed up request I can not check the X case.

ikidd:
Apparently commits take some time to propagate?  I pushed a commit about 13 hours ago and ended up reverting it because I thought it failed, but see it in the history now.  I just pushed another commit, no clue when it will show up in this PR.  If you don't see the latest commit I did just before this message, I haven't changed the code since the one this morning that eventually appeared.

In any case, I had the endstop reporting set up and tested fine pretty much at the beginning (i think it was in the first commit), but thought I'd just test it again before this last commit, seems to be working fine and independently.  As I mentioned before, the Y axis seems to work quite fine.  If I knock either sled back, it trues them up with no problems, driving the lagging sled while the front sled stops driving.

X just keeps driving both sleds until the lagging endstop reports (predictably enough, this bends the endstop lever on the lead sled).  If I swap the X and Y axis stepper/endstop pins in configuration.h, the problem follows the X axis to the previously working stepper pair, so it doesn't seem to be HW related.

Obviously I'm missing something in my X code, but I can't seem to find it.

As for using the second word, I'll happily move this to the second state2 once I'm satisfied that I've got X working correctly.  Just didn't want to introduce another possible source of error since state1 is known good.

Also curious why a fresh Dev copy exhibits software endstop behaviour.  I still have no facility to move Z below its value at initialization because I don't have HW endstops to home it against.  It does not behave like that in 0.92.9 Stable, in fact I copy my Stable config.h to the fresh Dev copy and it still behaves like it has SW endstops active.

Really wish there was a way to do breakpoints and debug stepping in Arduino.  I have to say, I respect the heck out of anyone that can build as complex a firmware as you have without having that ability.  It's quite frustrating to not be able to trace all the values in realtime to see where the issues lie.

ikidd:
Still don't see the last commit.  IDK what's going on here, I've sent Support a message asking why it's not showing up, and why the last one _eventually_ showed on the PR.  I might have to refork and push PR again if my repo is gonged.  

In any case, https://github.com/repetier/Repetier-Firmware/pull/668/commits/97713d6a8a0e56084fa36cd010d32b1641677bce from yesterday morning shows the code.

repetier:
Except a lot of chaos from your editor removing ending whitespaces so many lines differ now even if you did not change them everything seems fine on a good look.

From your description I take that X min endstop triggers and that motor stops and x2 min never triggers and so motor continues to move. So you should inspect everything regarding these parts starting from your configuration where you define it to every place where you use them. Add some Com::printFLN(PSTR("Msg"),value); for debugging or enable report endstop changes so you see if it gets triggered. If you see both got triggered it is the variable you test that gets not set/read correctly. Add a print in checkendstops when trigger gets detected. In sum it should allow to find the place where it gies wrong. Must be something dead simple like an extra char we always overlook.

ikidd:
Yes, I don't know why my editor is doing those whitespaces, I certainly don't touch those line.  Seems to do it for me in Atom and in Arduino IDE.

After I uploaded the last commit, I added a print of the multiXHomeFlags at each step (yah, I know, slows the printer right down) and enabled debug 64 for endstops and that.

It doesn't matter if either of the X endstops trigger, it continues to drive both motors.  Yet the flags are correct for the endstop state, if X_MIN trips first, flags moves to 2 and if X2_MIN triggers first, flags moves to 1.  But both motors continue to push.  So you'd figure it would only execute the one WRITE statement, but that's not the case on X.  It is on Y though.

If I swap X and Y stepper pins and endstop pins so the axis are fully swapped, it continues to fail on the new X and works on the new Y.

I've put the CNC away for a bit while I get some other work done, I'll keep an eye on this but probably won't work on it much until I get some vehicle repairs done.  Maybe I'll have an epiphany in the meantime.

repetier:
Ok, I think I have seen your error.  I think you have also set
DUAL_X_AXIS and ditto printing
and if you look into the code in that case it will execute both steps regardless of endstop flag. This flag is not for your case - it is to have 2 axis. Might also be something different. For testing add a return befor ethe else stopping your case so the following set high commands never get a chance to get executed, I think one of them does get executed making the problem.

ikidd:
I have those both turned off in the config.h.  Are you saying they'll execute as the code is written independent of the config setting?

I'll try a return there, it does make sort of sense given the lack of that code in the Ystep.

repetier:
It's at least the only chance I see the steps still get executed. I think in your debug the log messages disappear when endstop are triggered, right? So if it still moves on other part sets it instead.

ikidd:
Essentially I'll see 3333333 then one endstop goes High in the endstop report, then that flag (1 if it's X2, 2 of it's X1) until the lagging one triggers and endstops report X and X2 as High, then 33333 while they back up for the second touch and a final endstop report with both High.

That's the mystifing thing, it's throwing the right flags, but for some reason both X and X2 are stepping.  The logic doesn't seem like it should allow it, but it does.

The number of steps is huge so maybe the comms isn't keeping up, I might have to figure out a way to just log it to SPIFFs or something so I'm certain of the debugs.

Leitnin:
Just wanted to say I'm interested in this as well and will be looking at the code/testing on my MPCNC. Maybe together we can get this to 100%

Leitnin:
GOT IT!

Short answer: In Printer.h The #else after #if DUAL_X_AXIS is causing xstep and x2step to happen regardless of the logic in Multi_XENDSTOP_HOMING.

Long answer:
So, after a day and a half of digging through the code and failing over and over to even be able to debug (print statements lock everything up because of the frequency of steps so I slowed feedrate down to 2), I'm pretty sure I have found the issue.

I have NO IDEA how in the world to fix this on Github or to even add the changes to my fork or anything (I'm a mechanical guy not a software guy and all the "tutorials" for github might as well be greek as they assume you know 90% of everything already)

Here it is, in Printer.h, with indents to help show what's happening.


Up here is the code for multiaxishoming, and then comes: 

#if DUAL_X_AXIS
   #if FEATURE_DITTO_PRINTING
      if (Extruder::dittoMode)
      {
        WRITE(X_STEP_PIN, START_STEP_WITH_HIGH);
        WRITE(X2_STEP_PIN, START_STEP_WITH_HIGH);
        return;      
      }
   #endif
   if (Extruder::current->id)
   {
     WRITE(X2_STEP_PIN, START_STEP_WITH_HIGH);
   }
   else
   {
     WRITE(X_STEP_PIN, START_STEP_WITH_HIGH);
   }
#else      //   <-----THIS GUY IS THE LITTLE DEVIL as it means the next lines happen even if the multi_xendstop logic is right, which is why the flags are correct.
    
  WRITE(X_STEP_PIN, START_STEP_WITH_HIGH);   //This runs regardless of multi_Xendstop flags

  #if FEATURE_TWO_XSTEPPER
    WRITE(X2_STEP_PIN, START_STEP_WITH_HIGH); //This also runs regardless of multi_Xendstop flags
  #endif
#endif

Leitnin:
Two other things: Sorry  for the way I'm including these as text, but as I said I'm not up to speed on github yet.
__________________
1.
In Printer.cpp between line 1778 and 1779 there should be a setHoming(false);
This not being there is preventing the printer from doing the move that corresponds to ENDSTOP_X_BACK_ON_HOME = 1;

        PrintLine::moveRelativeDistanceInSteps(axisStepsPerMM[X_AXIS] * -ENDSTOP_X_BACK_MOVE * X_HOME_DIR,0,0,0,homingFeedrate[X_AXIS] / ENDSTOP_X_RETEST_REDUCTION_FACTOR, true, false);
        PrintLine::moveRelativeDistanceInSteps(axisStepsPerMM[X_AXIS] * 2 * ENDSTOP_X_BACK_MOVE * X_HOME_DIR,0,0,0,homingFeedrate[X_AXIS] / ENDSTOP_X_RETEST_REDUCTION_FACTOR, true, true);
        setHoming(false);  //ADD THIS GUY
_______________
2.
Two lines in Printer.cpp control the Y axis but reference ENDSTOP_**X**_RETEST_REDUCTION_FACTOR instead of Y

One line in Commands.cpp does the same for Z.

You can find them both easily by searching for ENDSTOP_X_RETEST_REDUCTION_FACTOR

Squadzone:
hello, i would like to inform, repetier now support dual endstop for Y axis and X axis, use Dev branch,  thanks to @repetier, i already test for both, and working great :)

repetier:
Closing this as we have a working solution now.

ikidd:
Sorry, do I read that as the multihoming is working now or we just have dual endstops as it had for a long time?

And that's interesting about the #ELSE executing no matter what, Leitnin.  It mystifies me why that would exec if you dont' have DUAL_X_AXIS flag set.

repetier:
Yes, the official branch for dev has now a working multi axis homing and you can even configure it in our config tool. You will then see a second endstop for selection if you have a mirrored motor.

PR #669, opened 2017-04-28
More Dual-X fixes + new RAISE_Z_ON_TOOLCHANGE configuration entry

Braccoz:
not much to say:

- fix for Y offsets bug with LAZY_DUAL_X_AXIS
- removed the blocking of Z_UP_AFTER_HOME for LAZY_DUAL_X_AXIS
- implemented new configuration entry RAISE_Z_ON_TOOLCHANGE which as name implies, raises Z before swapping extruder and then lowers it back down after it's done


repetier:
Ok have added it and modified a bit. If you do not use ifdef you need to define it always or it does not compile without being set. It's now also available in config tool in tool page.

Braccoz:
ok, roger that. it compiled for me though, weird. maybe not with a different arduino ide version

PR #673, opened 2017-05-14
Add support AZSMZ 12864 LCD

cxandy:


repetier:
Could you make such a patch against dev version? Development for 0.92 is stopped and all changes must go to dev branch. Also please do not include your personal configuration.h version so we can accept the patch.

PR #688, opened 2017-07-24
shared PWM budget for extruder and heating bed

mplatov:
This commit adds a compile time option to add shared PWM budget for heating bed and extruder, which could be useful when printer's power supply can't handle them both at full power at the same time (which is normally not needed anyway). If enabled, this "budget" allows to dynamically cap PID_MAX PWM value for the heating bed when extruder is also powered at the same (if extruder is not powered regular HEATED_BED_PID_MAX value is used)

### DIFF
--- a/src/ArduinoAVR/Repetier/Configuration.h
+++ b/src/ArduinoAVR/Repetier/Configuration.h
@@ -299,6 +299,20 @@
 #define HEATED_BED_PID_IGAIN   33
 #define HEATED_BED_PID_DGAIN 290
 #define HEATED_BED_PID_MAX 255
+
+/** Shared PWM budget for extruder and heating bed.
+Normally there is not need to run both extruder and heating bed at full power, so some printer 
+vendors may ship printers with underpowered power supplies that can't handle them both 
+simultaneusly. This option is useful for this particular case as it dynamically limits max power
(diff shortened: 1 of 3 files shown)


repetier:
I'm glad it works for you, but this model is missing the physical correctness as I described in my forum post. This makes it hard to understand how to handle it and also leads eventually to overcompensation as the power consumption is not taken into account correctly. Also I see some place for speed improvement to compensate for the extra float computations needed with my solution like running computation inside the heater loop. Extruders come first anyway and then the bed so we can sum up directly there.

Will add that feature once I get a bit more time.

mplatov:
It will be great if there will be more elegant solution in the firmware, however there is one more thing which is important to account for - stepper motors. Once I found (empirically) the "budget" when both extruder and heater can work simultaneously without crashing I tried moving the motors at the same time and of course it shut off again. Motors in my printer draw about 20W of power and I had to decrease "the budget" further quite a lot to make it work reliably even if motors are active, which is important because motors were are active when this problem happened in the first place.

repetier:
Yes, but you can control consumption to motors. That is why you need to say I have free X watt for temperature so after subtracting what motors can draw. Then you put the rest into bed if power is available.

PR #694, opened 2017-08-04
Fix typo in ui.cpp

Depaulicious:
Don't drink and `git` 😜


repetier:
Or better watch paste button while scrolling:-)

Have already fixed it locally and will be included in next update.
Thanks.

PR #697, opened 2017-08-06
MIXING_EXTRUDER (ui) show the default extruder

phd0:
show the default extruder then MIXING_EXTRUDER, not the ext1 one


repetier:
You are omitting case 0 in your solution. Correct solution is

#if NUM_EXTRUDER > 0
#if NUM_EXTRUDER == 1 || (NUM_EXTRUDER > 1 && MIXING_EXTRUDER > 0)
UI_PAGE4_T(ui_page3, UI_TEXT_PAGE_EXTRUDER_ID
#else
UI_PAGE4_T(ui_page3, UI_TEXT_PAGE_EXTRUDER1_ID
#endif
#else
UI_PAGE4_T(ui_page3
#endif


Will add it in dev version on next update as I see it is a nicer solution in this special case.

PR #719, opened 2017-12-19
Hide Extruder related menu if none extruder is defined

luc-github:
CNC or Laser machine may not have any extruder so displaying extruder related menu is confusing


repetier:
Looks good. Helps at least with pure cnc/laser solutions.

luc-github:
Yes hybride solutions won't be affected - for pure laser / cnc solution I think there is more to do as mentioned : https://github.com/repetier/Repetier-Firmware/issues/720
let me know if you agree and I can do another PR  to complete this one

PR #724, opened 2018-01-02
 Preliminary support for Trinamic TMC2130 drivers

alexsomesan:
This change introduces basic support for configuring TMC2130 drivers via SPI.
It has  so far only been tested on Arduino DUE + RADDS board + Watterott SilentStepSticks.
I'm waiting on additional stepper modules to arrive so I can test on Arduino Mega + RAMPS.

Sensorless homing IS possible (works for me) using this change, but works roughly.
Additional testing and tuning is required for smooth operation with sensorless homig.

I'm opening this PR to start the conversation around the implementation choices I've made so far.
Feedback and suggestions are highly appreciated.

Features (checked are completed):
- [x] Initializing drivers via SPI
- [x] Setting motor current over SPI via M906  GCode
- [x] Toggling StealthChop mode on and off over SPI via M915 GCode
- [x] Setting Stallguard2 threshold values over SPI via M914 GCode

@repetier Please have a look and let me know what you think so far. Thanks a lot!

### DIFF
--- a/src/ArduinoAVR/Repetier/Commands.cpp
+++ b/src/ArduinoAVR/Repetier/Commands.cpp
@@ -488,6 +488,88 @@ void motorCurrentControlInit() { //Initialize Motor Current
 }
 #endif
 
+#if STEPPER_CURRENT_CONTROL == CURRENT_CONTROL_TMC2130
+TMC2130Stepper* tmcDriverByIndex(uint8_t index) {
+    switch(index) {
+#if TMC2130_ON_X
+        case 0: // X Axis
+            return Printer::tmc_driver_x;
+        break;
+#endif
+#if TMC2130_ON_Y
+        case 1: // Y Axis
+            return Printer::tmc_driver_y;
+        break;
+#endif
+#if TMC2130_ON_Z
+        case 2: // Z Axis
+            return Printer::tmc_driver_z;
+        break;
+#endif
+#if TMC2130_ON_EXT0
+        case 3: // E0 Axis
+            return Printer::tmc_driver_e0;
+        break;
+#endif
+#if TMC2130_ON_EXT1
+        case 4: // E1 Axis
+            return Printer::tmc_driver_e1;
+        break;
+#endif
+#if TMC2130_ON_EXT2
+        case 5: // E2 Axis
+            return Printer::tmc_driver_e2;
+        break;
+#endif
+        default:
+            return NULL;
+        break;
+    }
+}
+
+void setMotorCurrent( uint8_t driver, uint16_t level ) {
+    TMC2130Stepper* tmc_driver = tmcDriverByIndex(driver);
+    if(tmc_driver) {
+        tmc_driver->rms_current(level);
+    }
+}
+
+void setMotorCurrentPercent( uint8_t channel, float level ) {
+    const uint16_t tmc_motor_current[] = MOTOR_CURRENT;
+    uint16_t raw_level = ( level * tmc_motor_current[channel] / 100 );
+    setMotorCurrent(channel, raw_level);
+}
+
+void motorCurrentControlInit() {
+    const uint16_t tmc_motor_current[] = MOTOR_CURRENT;
+    for(uint8_t i = 0; i < (sizeof(tmc_motor_current) / sizeof(uint16_t)); i++) {
+        setMotorCurrent(i, tmc_motor_current[i]);
+    }
+}
+
+void motorCurrentReadings() {
+    Com::printF(Com::tTrinamicMotorCurrent);
+#if TMC2130_ON_X
+    Com::printF(Com::tSpaceXColon, (uint32_t)(Printer::tmc_driver_x->rms_current()));
+#endif
+#if TMC2130_ON_Y
+    Com::printF(Com::tSpaceYColon, (uint32_t)(Printer::tmc_driver_y->rms_current()));
+#endif
+#if TMC2130_ON_Z
+    Com::printF(Com::tSpaceZColon, (uint32_t)(Printer::tmc_driver_z->rms_current()));
+#endif
+#if TMC2130_ON_EXT0
+    Com::printF(Com::tSpaceEColon, (uint32_t)(Printer::tmc_driver_e0->rms_current()));
+#endif
+#if TMC2130_ON_EXT1
+    Com::printF(PSTR(" E1:"), (uint32_t)(Printer::tmc_driver_e1->rms_current()));
+#endif
+#if TMC2130_ON_EXT2
+    Com::printF(PSTR(" E2:"), (uint32_t)(Printer::tmc_driver_e2->rms_current()));
+#endif
+    Com::printFLN(Com::tSpace);
+}
+#endif // CURRENT_CONTROL_TMC2130
 
 #if STEPPER_CURRENT_CONTROL == CURRENT_CONTROL_MCP4728
 uint8_t   _intVref[]     = {MCP4728_VREF, MCP4728_VREF, MCP4728_VREF, MCP4728_VREF};
@@ -608,6 +690,43 @@ void motorCurrentControlInit() { //Initialize MCP4728 Motor Current
 }
 #endif
 
+#if defined(DRV_TMC2130)
+void microstepMode(uint8_t driver, uint16_t stepping_mode) {
+    TMC2130Stepper* tmc_driver;
+    if(tmc_driver = tmcDriverByIndex(driver))
+        tmc_driver->microsteps(stepping_mode);
+}
+
+void microstepInit() {
+    const uint16_t microstep_modes[] = MICROSTEP_MODES;
+    for(uint8_t i = 0; i < (sizeof(microstep_modes) / sizeof(uint16_t)); i++) {
+        microstepMode(i, microstep_modes[i]);
+    }
+}
+
+void microstepReadings() {
+    Com::printF(Com::tTrinamicMicrostepMode);
+#if TMC2130_ON_X
+    Com::printF(Com::tSpaceXColon, (uint32_t)(Printer::tmc_driver_x->microsteps()));
+#endif
+#if TMC2130_ON_Y
+    Com::printF(Com::tSpaceYColon, (uint32_t)(Printer::tmc_driver_y->microsteps()));
+#endif
+#if TMC2130_ON_Z
+    Com::printF(Com::tSpaceZColon, (uint32_t)(Printer::tmc_driver_z->microsteps()));
+#endif
+#if TMC2130_ON_EXT0
+    Com::printF(Com::tSpaceEColon, (uint32_t)(Printer::tmc_driver_e0->microsteps()));
+#endif
+#if TMC2130_ON_EXT1
+    Com::printF(PSTR(" E1:"), (uint32_t)(Printer::tmc_driver_e1->microsteps()));
+#endif
+#if TMC2130_ON_EXT2
+    Com::printF(PSTR(" E1:"), (uint32_t)(Printer::tmc_driver_e2->microsteps()));
+#endif
+    Com::printFLN(Com::tSpace);
+}
+#else
 #if defined(X_MS1_PIN) && X_MS1_PIN > -1
 void microstepMS(uint8_t driver, int8_t ms1, int8_t ms2) {
     if(ms1 > -1) switch(driver) {
@@ -760,7 +879,7 @@ void microstepInit() {
     for(int i = 0; i <= 4; i++) microstepMode(i, microstep_modes[i]);
 #endif
 }
-
+#endif
 /**
 \brief Execute the Arc command stored in com.
 */
@@ -2287,13 +2406,14 @@ void Commands::processMCode(GCode *com) {
         break;
 #endif // FEATURE_SERVO
     case 350: { // M350 Set micro stepping mode. Warning: Steps per unit remains unchanged. S code sets stepping mode for all drivers.
-#if defined(X_MS1_PIN) && X_MS1_PIN > -1
+#if (defined(X_MS1_PIN) && X_MS1_PIN > -1) || defined(DRV_TMC2130)
         if(com->hasS()) for(int i = 0; i <= 4; i++) microstepMode(i, com->S);
         if(com->hasX()) microstepMode(0, (uint8_t)com->X);
         if(com->hasY()) microstepMode(1, (uint8_t)com->Y);
         if(com->hasZ()) microstepMode(2, (uint8_t)com->Z);
         if(com->hasE()) microstepMode(3, (uint8_t)com->E);
-        if(com->hasP()) microstepMode(4, com->P); // Original B but is not supported here
+        if(com->hasP()) microstepMode(4, (uint8_t)com->P); // Original B but is not supported here
+        if(com->hasR()) microstepMode(5, (uint8_t)com->R);
         microstepReadings();
 #endif
     }
@@ -2551,8 +2671,12 @@ void Commands::processMCode(GCode *com) {
     case 908: { // M908 Control digital trimpot directly.
 #if STEPPER_CURRENT_CONTROL != CURRENT_CONTROL_MANUAL
         uint8_t channel, current;
-        if(com->hasP() && com->hasS())
+        if(com->hasP() && com->hasS()) {
             setMotorCurrent((uint8_t)com->P, (unsigned int)com->S);
+        }
+#if STEPPER_CURRENT_CONTROL == CURRENT_CONTROL_TMC2130
+        motorCurrentReadings();
+#endif
 #endif
     }
     break;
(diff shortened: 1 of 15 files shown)


repetier:
First thanks of taking care of the TMC2130 driver problem (meaning not existing support).

First, please do pull requests always against the dev version, also at the moment both are identical master is seen directly and dev only when I merge it into master after a while.

For M908 you should better add a own setMotorCurrent function just as the 2 already existing and not put everything in the m code selector.

For defining microstepping use microstepMode alternative.

Can you say how stallguard works? Does it have a extra signal line for detecting it? Does it work reliable? I saw you added some code to homing but endstops are untouched so I guess that is not finished? I think here it might need a switch for each endstop if it is not 100% exact especially for Z axis homing accuracy is important.

As you might have seen, it is quite hard to integrate it into current firmware. That is why I'm currently working on a new configuration design breaking these things into modules. Hope to get a first version published soon. There I have a stepper module, so each stepper can use different drivers making everything more flexible. Current config looks like this;

class StepperDriverBase {
public:
    StepperDriverBase(EndstopDriver *minES, EndstopDriver *maxES):minEndstop(minES),maxEndstop(maxES),direction(true),position(0) {}
    virtual ~StepperDriverBase() {}
    EndstopDriver *getMinEndstop();
    EndstopDriver *getMaxEndstop();
    /// Allows initialization of driver e.g. current, microsteps
    virtual void init() {}
    /// Executes the step if endstop is not triggered. Return tru eif endstop is triggered
    virtual bool stepCond() = 0;
    /// Always executes the step
    virtual void step() = 0;
    /// Set step signal low
    virtual void unstep() = 0;
    /// Set direction, true = max direction
    virtual void dir(bool d) = 0;
    /// Enable motor driver
    virtual void enable() = 0;
    /// Disable motor driver
    virtual void disable() = 0;
    // Return true if setting microsteps is supported
    virtual bool implementSetMicrosteps() {return false;}
    // Return true if setting current in software is supported
    virtual bool implementSetMaxCurrent() {return false;}
    /// Set microsteps. Must be a power of 2.
    virtual void setMicrosteps(int microsteps) {}
    /// Set max current as range 0..255
    virtual void setMaxCurrent(int max) {}
    EndstopDriver *minEndstop;
    EndstopDriver *maxEndstop;
    bool direction;
    uint32_t position;
};

As you can see it has already slots for init, microsteps and current. It also associates the endstops with the stepper driver making it a breeze to change between normal endstops and endstops controlled by stallguard. You could even mix it.

Since we stopped development on V1.0 and only want to add new drivers/boards and bug fixes I would suggest to add drivers only with current/microstep addition. No eeprom (also good idea for V2) and no ui.

What printer type do you have? I'm currently working on making pure cartesian printers work first with the new system. So I'd love to see you also add the functionality into a V2 driver, which will be much easier once we have generalized the driver interface required. But I think that will be a good test if all required functions are implemented.

alexsomesan:
Thanks for the super-fast review!

First of all, the new driver abstraction class looks awesome!
I'm really looking forward to that new framework for drivers and I will definitely port the TMC2130 support as soon as you merge it in the dev branch.

Now about this PR:
- I agree we can leave out eeprom and ui for V1.0 - I can add them as soon as V2 takes shape
-

alexsomesan:
Thanks for the super-fast review!

First of all, the new driver abstraction class looks awesome!
I'm really looking forward to that new framework for drivers and I will definitely port the TMC2130 support as soon as you merge it in the dev branch.

Now about this PR:
- I agree we can leave out eeprom and ui for V1.0 - I can add them as soon as V2 takes shape
- I used the Marlin M-Codes (for compatibility) since there were no equivalents for these operations in Repetier.
Because of that, the current setting is on `M906` right now. If you prefer, I can move it to the existing M908 and just add another implementation of `setMotorCurrent` for TMC, like you suggested.
- Thanks for telling me about `microstepMode` - I will make use of it like you suggest.

Stallguard, in simple terms, works by sending a pulse on the DIAG0 or DIAG1 pins (configurable) of the driver chip when the motor is stalled. These pins are simply connected to the normal endstop headers on the board, like electronic endstops.
They are a usually on the order of milliseconds long so the usual endstop code should be able to pick them up. 
That is why I made no changes to the endstop code so far. Seems to work on the DUE which is quite fast, but I haven't yet tested on AVR.

Stallguard is only reliable within a specific range of speeds. It needs a few parameters tuned correctly to trigger reliably. I am still trying out these parameters to see what works best. I'll update the PR as I find better ones.
However, they depend a lot on each printer hardware (friction of rails, motors). So ideally to make the feature completely user friendly we would have an auto-tuning process like Trinamic describes in their datasheet. That's something for later, when basic support is stabilized, I think.

I don't expect anyone to use stallguard for Z-min homing right now. Stallguard hasn't yet proven to be accurate enough for that. I think a z-probe is still required in that case, especially for bed-leveling.
However, with Stallguard available, there is the option of "homing" an axis in both directions (min and max) which could set the existing software endstops values without requiring the user to provide them. I think that could be something to explore for X, Y and Z-max.

alexsomesan:
Forgot to mention my setup:
I'm testing this on a Hypercube Evolution CoreXY printer, with 0.9 degree motors.
The electronics are a RADDS board on top of a DUE.
I can also test on my other Prusa clone, for the covering Cartesian case.
I don't have a Delta though, so I can't test for that case.

repetier:
Ok, so if understand it correctly you can not say in which direction stall occurs, so you just set it for the endstop you test anyway. And it vanishes on it's own after a short time, so you need to test it in time. For homing both is no problem and the min speed can also be solved using proper homing speeds. 

V2 will have cartesian system first, but core xy should be very easy to add if that works, so also be testable fast. Only at the beginning config will change frequently until all parts are adopted to the new module system. I'm just starting with motion because I also want S-Shape acceleration for smoother prints:-) I think this will be an awesome update once we are finished making much more things possible then before.

alexsomesan:
Oh yeah! S-shape accel will be a killer! Really looking forward to that! 
I was wondering when will firmwares start exploring that, but I guess with AVR being the prevalent platform until now, it just wasn't powerful enough for the math. Are you computing the curves on the fly or using some sort of pre-defined lookup tables?

You understand correctly - the same stall signal is used in both directions. It actually makes sense - you can only ever hit one endstop when moving, not both at the same time. I guess the only reason mechanical stops are used in pairs is because they need to be physically placed in different locations. But virtual endstops, like Stallguard don't have that requirement. However, you can even poll the value of the direction pin on each driver via SPI if you need to know it.

repetier:
I compute s-curve on the fly. Might even work on avr if it is not a delta that already has so many extra calculations.

In V2 only the endstop in move direction is tested, so it would assign it correctly to the right one during move. Problem might only arise if we want to move back after e.g. homing and it is still triggered. We have to check that case then.

alexsomesan:
I managed to get Stallguard tuned nicely and homing reliability on my CoreXY.
Cartesian should work just well. I’ll test that next.

I’ll make video demo tomorrow and update the PR with the tuning and the changes we discussed earlier.

Really happy that it works well!

boelle:
@alexsomesan cant wait to see the demo video

@repetier will this mean that the online config will get a box where we select what stepper drivers we use? and if we select the tmc2130 we will get box's where we select current and microsteps etc?

alexsomesan:
Demo video of what these change do:
https://www.youtube.com/watch?v=8wQiilALDm4

alexsomesan:
@boelle Z-min homing is not in the scope of this change. It seems that most who have tried it noted that StallGuard is not precise enough for Z-min homing and bed leveling. A probe seems to be superior at that for the time being.
I'll still give it a try but only after basic support for configuring the drivers and X / Y homing is stable.

repetier:
For 1.0 there will be only a config option that you have TMC2130 for all steppers or not. Config is not so fina granular that we can do it more precise. There is already an array for default current and microsteps e.g. for RAMBO board.
//  Microstepping mod eof your RAMO board
#define MICROSTEP_MODES { 8,8,8,8,8 } // [1,2,4,8,16]
// Motor Current setting (Only functional when motor driver current ref pins are connected to a digital trimpot on supported boards)
#define MOTOR_CURRENT_PERCENT { 55,55,55,55,55 }

but you need to edit it afterwards, no config for that now. It is a bit awkward since it is only available for some boards and then even depends on implementation and number of drivers supported.

In V2 you have to configure each stepper driver as well. That will make it easier also to mix drivers and do other fancy stuff. That is one reason we want to change the configuration scheme.

alexsomesan:
I updated the PR to make use of M350 for setting microsteps and M908 for setting currents, like you suggested.
I turned out quite messy, like you said in the beginning. On the good side, it will clean up nicely with the new StepperDriver classes of V2. All this mess can be nicely abstracted then.

Can you please have another look and let know if anything else needs touching up?
@repetier

repetier:
Looks very good so far and also messy as you said. Did you have a look here:
http://reprap.org/wiki/Gcodes#M914:_Set.2FGet_Expansion_Voltage_Level_Translator
M914: Set/Get Expansion Voltage Level Translator
M915: Configure motor stall detection

So you give them a new meaning compared to RepRapFirmware. Are your codes the same as for marlin and they have not documented it in the wiki?

Config should also be for EXT1 and EXT2 since you enable use of them.

For V1 I not feel like wanting to add all these configs especially since they are stepper socket based, so I will just add a all or none switch here and just add the bulk of configs. They can overwrite them in manual additions then if they do not like it for a special axis.

There are a few things I do not understand that would be great to know for V2 so I can add support for this more easily:

+#define TMC2130_STEALTHCHOP_X       1  // Enable extremely quiet stepping
 +#define TMC2130_INTERPOLATE_256_X   true
 +#define TMC2130_STALLGUARD_X        0
 +#define TMC2130_PWM_AMPL_X        255
 +#define TMC2130_PWM_GRAD_X          1
 +#define TMC2130_PWM_AUTOSCALE_X  true
 +#define TMC2130_PWM_FREQ_X          2

What does that all mean and where is the microstepping 1/2/4/8/16? Imagine you are a normal user and have no idea about stepper drivers and how they work (just like me). What should they set, what is the allowed range of values, what are they fore.

For V2 I need a set of extra virtual functions to allow changing them over the general interface. I think stealthchop only needs to be set in configuration. Normally you want it quiet or not.
I want to add a microstep function with mircosteps. One for current in range 0-65535 whcih can then be scaled down to motor/chip units.
Then there is this PWM handling I currentyl do not have an idea what it is for and if we can group them. Any hints here are welcome.

Last question is headless homing. As I understand it, you enable the "endstop" handling in homing start and disable it at end of homing. Normal endstops are activated all the time and also tested all the time. Is there a reason to not just simply enable them at startup? Only reason i see is false signals during print would happen frequently so it would be better to only use them if they are REALLY required since normal moves should not hit sides anyway. 

Except pins.h all files are identical with AVR version and you can simply copy the files the avr folder.

alexsomesan:
That was a very fast review, thank you so much!

First, the G-Codes. Since these are the "UI" of the firmware, it's important to get them right.
Thanks a lot for pointing out the RepRapFirmware wiki. I admit I missed the fact that they already document G-Code specific for stall detection.
My approach, for lack of a better convention, was to stay compatible with Marlin g-codes, since they had these features first. What is the general convention for Repetier? Do you try to follow RepRapFirmware as a standard?
I  will change the behavior of M915 to follow what the wiki describes(it's even more flexible than what I have).
For M914, which toggles the "quiet mode" on or off per driver, I can't find an equivalent in the wiki. Marlin also uses this for the same purpose, so that's how I picked it. Do you have a suggestion for another one?

> For V1 I not feel like wanting to add all these configs especially since they are stepper socket based, so I will just add a all or none switch here and just add the bulk of configs. They can overwrite them in manual additions then if they do not like it for a special axis.

When you say this, do you mean configure all axes using a single value per parameter in Configuration.h?
We can certainly do that, except for TMC2130_STALLGUARD_ which has to be configurable per axis because this adjusts the sensitivity for stallguard when homing and different axes have different mechanical resistance.

> What does that all mean and where is the microstepping 1/2/4/8/16?
Microstepping is read from the existing MICROSTEP_MODES macro, like you asked me earlier (what M350 uses).

We can drop TMC2130_INTERPOLATE_256_ since it's likely going to be true all the time for 3D printers.
It enables interpolation of microstepping to 256 internally in the driver. The chip internally multiplies the STEP signals to 256 resolution. Makes for smoother motion.

> For V2 I need a set of extra virtual functions to allow changing them over the general interface. I think stealthchop only needs to be set in configuration. Normally you want it quiet or not.

Is this something you want in this PR or should I just expand your example base class in a separate document?

> Then there is this PWM handling I currently do not have an idea what it is for and if we can group them. Any hints here are welcome.

These are advanced configuration for fine-tunnig the stealhChop (quiet) mode. They are described in the TMC2130 datasheet, in chapter 6.
We can drop them for now since they require an elaborate tuning with oscilloscopes which most users will not do. It works well enough with the defaults that the driver sets.

> Last question is headless homing. As I understand it, you enable the "endstop" handling in homing start and disable it at end of homing.

StallGuard (signaling the endstop condition) only works in the louder driver mode called SpreadCycle. If the user configures the drivers to run in quiet mode (most often case) then this has to be temporarily turned off during homing. I turn it back on at the end of the homing function and restore the "normal" operating values for the flag I had to change for homing. That is the purpose of the code at the beginning  and end of the homing functions.

If the printer is running in quiet mode, it will not detect stalls during normal printing. Thus no endstop signal will be sent in quite mode even if ALWAYS_CHECK_ENDSTOPS is on. This is a limitation of the Trinamic driver.

False positives can be resolved by choosing the right sensitivity value for TMC2130_STALLGUARD_, but the actual values depend on the mechanical resistance of each printer, so users need to determine them. This can be automated in the long term by implementing the tuning process described in the datasheet, but that's quite a bit of work too.

(See the prusa demos - they only do the lost step recovery in the "loud" mode).

repetier:
No I do and can not follow RepRapFirmware. They often use non gcode syntax for gcode which repetier can not handle due to binary protocol. So I try to be compatible where possible and otherwise have my own solution.

I try to reuse existing m codes where possible, so if you say marlin uses them already, it is ok to use them also.

Regarding configuration I would leave what you already have. What I mean is I add only the driver selection and endstop handling as option and if selected put all your configs into it. They can then override in in manual additions field if they want different values for all the 30 parameter remaining. Guess the defaults are normally good enough and experts can still tweak them. In V2 we need configs per module implementation, so we can then make these per driver anyway.

V2 stuff is not for this PR. It is information for me so I can add the required functions in stepper base as you can only implement what I allow in stepper base. So if we say set current is a function I add it as dummy virtual function and drivers being able to do so would then implement the function.

PWM settings - leave them but do not offer software adjustments. Make a note that this is explained in Chapter 6 and only for real experts to touch. Then no one will ask and be happy they had an expert predefining them.

Endstop handling is understood. Will add also virtual function for start and stop homing then in V2.

So to summarize what we need in V2
- setCurrent 0-65535
- setMicrosteps 0-255 
- setStealthMode 0-255 in case more then 1 is available
- startHoming
- endHoming
- positionMask 0xffffffff or 0xfffffffe ??

I heard the enhanced microstepping mode is better when step is 50% on or off. You can set driver to step on changing signal on only low to high. We can do that in V2 by making unstep a empty function and toggle signal on step. I'm not sure what will happen if we change direction (which is done before). Would changing direction after rising high cause a step or not. I think not would be the right solution, then we do not need to mask position to always have equal step count before changing direction.

Did I miss a function needed for handling all extra function?

alexsomesan:
Thanks a lot for the confidence, @repetier !

I will follow up today with a small change to simplify the amount of parameters needed in Configuration.h (like you suggested in the last comment). That's the last outstanding issue.

repetier:
I'm just merging and modifying config hints. Please wait and see what I do.

repetier:
Ok updated dev version. Also added a note that this library

https://github.com/teemuatlut/TMC2130Stepper

is required. Is that correct?

Please use merged version for further improvements.

alexsomesan:
Yes, that library is needed. It's available directly from the Arduino library manager.

I already deleted the branch for this PR. I'll create new ones from dev when I need to make changes.
I'll wait for your changes then.

Also, is the configuration tool's sourcecode available anywhere? I could also try adding controls there for setting the Trinamic parameters if you agree.

In the mean time I am thinking about what is necessary in the V2 base class. I'll send you my suggestions over the weekend.

Thanks again!

repetier:
Sorry, config tool is not open source. Do not want anyone to change it as that is what users use mostly and I need control over this, so I need to do this:-( Will add support for it once we are done here in dev version.

V2 is in the meanwhile starting to work. Could already move around with new system, so now I need to start homing which is the next difficult thing to make general.

repetier:
Ok, config tool for 1.0.1dev is now online with trinamic driver. Let me know if you have some more informations I could provide to make live easier. For the timings e.g. what does 300 mean? Is that 300 mm/min or full steps/s?

alexsomesan:
That’s great news!

I’m trying it out now. I found some stuff that needs a bit of tweaking, I
will open issues for each so you can track them.
For example, they should be allowed also for CoreXY. That works too and is
what I’m testing with.

The duration and speed parameters are a bit tricky to explain to users.
They all relate to the TStep value which is the duration of a native
microstep (after interpolation is applied). This duration is reported in
real-time by the drivers in a register and can be read back. All timing
values are a multiple of this TStep. But it’s not available or usefull to
users.

I think the configuration tool should just have sets of known good values
for various hardware configurations (like defaults for Cartesian, for
CoreXY, etc).

The Trinamic drivers are really meant to be tuned at runtime. What we
really need to support them correctly is a set of “calibration” routines
that would be run once like the bed leveling, tune all the necessary
parameters and save them to eeprom. It seems this is the right way to
implement them.

I’m planning to work on those calibrations for v2.

boelle:
side Q... would it at all make sense to add the TMC2100 ?

config can be hardwired or done by 3 digital lines and each driver has 2 diagnostic outputs

alexsomesan:
@boelle I think the TMC2100 can be fully used right now without any special software support.
They also don't implement StallGuard so they cannot be used for sensorless homing. The diagnostic pins are not that interesting because they only signal error states like shorted motor coils or overtemperature condition.

You should be able to use them with the current Repetier stable.
What feature are you missing to use the TMC2100 ?

boelle:
@alexsomesan it was only a "half cooked" thought so to speak

you are right that we dont need anything special to use them

but if they should be added to the mix maybe changing the setup in firmware instead of chaning the CFG pins

and maybe use the over temp detect to figure out how fast they can move without getting to hot?

this is just thought thou and i'm ready to stand corrected etc :-D

alexsomesan:
Those are all good ideas, I agree.
They would be easier to implement in the re-designed driver architecture
that Repetier is working on for V2. We should wait for that first as it's
going to really make it simpler to support new drivers.

The idea behind this PR was to unblock the path for TMC2130 which were
otherwise not usable at all.
As you can see from the conversation above, the implementation is very
basic right now and only meant to get things going.
The really nice features are going to come in V2.

repetier:
You can configure it with any printer type, but I think you mean the comment for sensorless homing. Have added core as well to make clear that this is still cartesian:-)

Timing are in deed hard to understand. We leave them as they are so they can be tweaked, but I have marked them as expert settings anyway so most users will just leave it as it is anyway. If you have some better comments or hints let me know and I extend the description, no problem.

V2 is making good progress. Could already sucessfully print as my printer has accidetially exactly the parts supported:-)  Working now on analog input, then I can start making extruders with new system and then I'm through first pass and can start adding additional alternative modules for all the variants users want. So soon you can start porting the driver.

alexsomesan:
@repetier I also develop on a RADDS + Due. What other parts are needed to have go at it?
If I want to try it out, I would just build what's in the dev2 branch? 
I want to start experimenting with porting these features to V2.

boelle:
@repetier 

i have an analog touch sensor (https://miscsolutions.wordpress.com/mini-height-sensor-board/)

how easy would it be to add that one in V2?

should i open another feature request for this?

repetier:
@boelle In V2 it would get possible. V1 is very slow on analog input as it averages very often. With the new system we can select per analog input if we want to average at all and how often. So with some simple new modules it can be converted in a digital input, only drawback is that you need to use low speed as we can not measure per step but with fixed frequency (ca. 4khz) so speed must be < 4000 steps per second for probing to be quite precise.

@alexsomesan Config is for felix printer (radds compatible) with 2 extruders and heated bed. LCD is not working so disabled. Set motherboard to 402 and just upload and it should work out of the box, but as cartesian (core xyz module is missing atm). But that just means moves are diagonal. core module will be also quite simple and similar to cartesian. I just make one for you so you can also tell me if it works:-) Configuration is of course a mess as it is what I'm currently changing so it uses parts from old and new configuration (configuration_io.h and confugration.h are now used together as one config). You need visual studio code as described in readme. And then just check out V2, see and compile. All you need to touch at first is src/drivers/stepper.h and add the TMC2130 variant. Later I will also add multiplex drivers and mixing drivers to use multiple motors per axis. But they will always just use the base interface so it does not matter which type of driver is used then as input.

repetier:
Ok, have added core xyz math, just select printer type 2 in configuration.h and set factors accordingly.

Motors are configured in configuration_io.h - see that I use a inverted output pin to invert if low should be high.

repetier:
Would be interesting if homing works. I think it does not because I test a sensor only if the motor moves in that direction but for corexyz that does not match, so it could be that you have to live without homing at the moment until I have a solution for that.

PR #727, opened 2018-01-08
Simpler configuration for TMC2130

alexsomesan:
This PR is to propose a bit of reorganizing for the Trinamic code in order to reduce the number of user-facing parameters in Configuration.h

It includes:
 - added global paramters (affect all drivers)
 - isolated overrides and validations into new header Trinamic.h
 - some extra explanations in comments around configuration parameters

@repetier


repetier:
Ok have included the improvements. Makes it easier for config tool.
Do you think you are now done or should I wait with adding to config tool? If you think it is perfect I will add it maybe at the weekend.

alexsomesan:
I think for V1 this should be enough to get people going. 
I'd rather help out with V2 support next and implement any improvements after that.
Would be great to get config tool support for this part. Thank you!

PR #766, opened 2018-03-25
SMART RAMPS with EEPROM motherboard fixes

lorf:
Small fixes for the SMART RAMPS with EEPROM board (ID 413)

1st patch fixes double-defined TWI_CLOCK_FREQ when MB 413 selected.

2nd patch is an addition to commit 9072c95.


repetier:
The changes are included in latest dev version. Also added the TWI to 408 so it has no doubles any more. This patch is against master version and we do not patch here. They always go to dev first and after some time all patches are bundled in a bigger update of master.

PR #771, opened 2018-04-08
 Allow build of ArduinoAVR using only avr-gcc - Fix Makefile so it actually builds

rickyrockrat:
This may be more than needed, but this also includes the needed library files from the Arduino IDE, which should be GPL2.

I'm sure there is a better place for the libraries than were they are, and perhaps a readme of how to re-create the arduino.lib directory would be better. There's comments in the Makefile on how I created that directory.

This was built on Devuan Jessie.

Seems like an awful lot of work to just get a few files to compile, and this will save others used to doing embedded work a quick, familiar solution.


repetier:
Also I like the fact that the makefile gets an update, also it seems nobody is using it (never got a question about this not working any more), I do not really like all these other files in the directory.

Why did you not just use a variable pointing to the Arduino IDE with all the files. That would also make updates easier as you just reference the updated files instead.

BTW: Does the makefile also work with the new src directory containing the SdFat library? That was added yesterday, so might not be included.

PR #773, opened 2018-04-08
Allow build of ArduinoAVR using only avr-gcc - Fix Makefile so it actually builds #3

rickyrockrat:
So third time hopefully is the charm. I've added a script to pull the libraries from a truncated Arduino repo I built that mirrors the Mother Ship, complete with the correct tag.

This was built on Devuan Jessie.

Seems like an awful lot of work to just get a few files to compile, and this will save others used to doing embedded work a quick, familiar solution.

### DIFF
--- a/src/ArduinoAVR/Repetier/.gitignore
+++ b/src/ArduinoAVR/Repetier/.gitignore
@@ -1,0 +1,2 @@
+applet
+arduino.lib
--- a/src/ArduinoAVR/Repetier/.gitignore
+++ b/src/ArduinoAVR/Repetier/.gitignore
@@ -1,2 +1,3 @@
 applet
 arduino.lib
+ard.all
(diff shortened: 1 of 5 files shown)


repetier:
Yes, that is much better. Only thing I'm not sure with is why adding fake_new_.. files. These will get compiled if I compile using arduino and I know I do not need them. Did you forget to delete them or is there a special reason and how does it affect regular compilation?

rickyrockrat:
How does arduino know to complie them? Does it just blindly compile those files? If so, where can I put them that Arduino won't find them.

rickyrockrat:
OK, NOW maybe Arduino won't complain. The files only get created when you run this make file...unless you are telling me that Arduino is calling this make file? I obviously know nothing about Arduino.

PR #776, opened 2018-04-10
Update makefile subsystem for ArduinoAVR stand alone make for all versions of Arduino

rickyrockrat:
This is an update to the last push that now supports both old and new avr-gcc compilers, and supports 1.0.5 through 1.8.4 versions of Arduino, and updates the sparse Arduino clone URL.


repetier:
Ok have included it.

PR #792, opened 2018-05-27
Update pins.h

rafaljot:
there is no SD on RAMPS-FD
  #define SDCARDDETECTINVERTED false 
generates only unnecessary warnings because it is repeated in configuration.h


repetier:
I have changed the upcoming updates to contain a undef to prevent the error messages instead. This also helps in other cases not covered in your solution. But thanks for pointing to the problem.

PR #812, opened 2018-08-19
Add an option for median instead of mean for bedleveling repetitions

sjoerdtimmer:
Useful for piezo probes.

I have under-bed piezo probes for automatic bed corrections. These probes are extremely accurate but also extremely sensitive. Sometimes they pick up a false positive. In that case all measurement are very close together except the false trigger which is much higher. Taking the mean reduces the error but taking the median completely removes it.


repetier:
Good idea for the problem. I put a optimized version into the dev tree, so it comes into master once 1.0.3 gets released officially. We do not make merges against master.

Main difference is that the new version does not use malloc/free any more. That should be completely avoided if somehow possible and in this case we know the maximum repetition count so we can simply use a fixed sized array:-)

And I use it as definite define with 0 or 1 set. I prefer this a sit makes things easier.

PR #815, opened 2018-08-24
Improve Wanhao I3 V2.1 support

Squall-DA:
Modify the Melzi configuration so that the software compiles for the Wanhao duplicator I3 V2.1. If my understanding is correct the issues that affect the DI3 should affect anyone who pairs the RepRap discount display with the Melzi Board. This should fix those issues.


repetier:
Thanks. Have merged it into the dev version.

PR #830, opened 2018-11-05
Add support for RURAMPS4D v1.3 board

Braccoz:
Add support for RURAMPS4D v1.3 board


repetier:
Thanks for the extra settings. Will soon also release it in the config tool.

PR #834, opened 2018-11-29
Update to master

badvision:


repetier:
I guess you did not mean to modify official version to your personal settings? Or does this also contain any fixes that I should import. Closing for now as it is not against dev and seems to be a personal fork.

badvision:
Oh no I meant to PR the other way, sorry about that!!

PR #835, opened 2018-12-14
Remove IDE files from codebase and mark them ignored in .gitignore

Cabalist:
Hey there,

This change removes the `.idea/` directory used by Jetbrains products.  This was causing issues in my my IDE (IntelliJ on macOS) when I opened it because some paths were different.   I expect this was accidentally checked in at some point and I've added it to the `.gitignore` as well.    

This should have no change to the firmware or to any existing dev environments. 

Thank you for all your work!


Cabalist:
Noted.  Thanks!

PR #838, opened 2018-12-29
fix: compilation errors (assignment instead of comparison)

redrathnure:
kind of typo


repetier:
Thanks. Will be fixed by a update coming hopefully today.

PR #841, opened 2019-01-05
i need help

Thercchannel:
I installed the community mod master 1.4..! But my Printer is going when i start printing in the home modus as normal. after that he starts the print! And afte 2 minutes the printer goes to the x and y offset of extruder and bed! then he prints contiune. He make a line over the heatbed. Is that a new trick and is that normal?
 Sorry for my bad English guys.


repetier:
Please to nor run pull request for questions. Especially not if the sources do not come from our repository and are completely different.

If it is normal depends on the gcode you have send. If you first home and then wait for target temperatures there will be a long delay before print continues, so that can be ok.

PR #843, opened 2019-01-11
fix complie error when disable Russian language.

hfutxqd:


repetier:
Thanks for information. Have checked it and it was in deed wrong, but in your solution you also have removed one wrong #endif instead of the right spot where it was too much. Have updated how I think it should be.

PR #855, opened 2019-03-14
fix SDCard on RuRamps4D v1.3

Braccoz:
small fix for SD card on ruramps 1.3


repetier:
Thanks for the correction.

PR #859, opened 2019-04-09
Add "threshold" to tmc2130 stallguard desc

ZibbeZabbe:
Simply adding the word threshold to TMC2130_STALLGUARD  description helps clarify what the numbers do


repetier:
Thanks. Will be put in next dev version and later on master.

PR #862, opened 2019-04-23
M118 message to host

tecan:
for sending messages to host , need this for https://github.com/tecan/QPrint3D project


repetier:
I will add it in the next dev version as update with 2 changes:
1. Will add Echo: before output as state on reprap gcode description. 
2. Not sure why you want to stop parsing text at first space like M117. Don't think that is a good idea. Only * should stop it as that is start of checksum.

PR #866, opened 2019-05-07
Update pins.h

RAyWB:
Typo


repetier:
Added in last commit.

PR #869, opened 2019-05-10
Patch 4

RAyWB:
change pin numbers to match AGC pinout


RAyWB:
> 
> 
> change pin numbers to match AGC pinout

sorry, wrong branch

PR #874, opened 2019-05-28
Support for Fysetc Mini12864 Panel v2.1.

CyberDeck:
This pull requests adds support for the **Fysetc Mini12864 Panel v2.1** for the **Fysetc F6** board.
Motherboard was included in the past, but not the Feature_Controller.

This LCD comes in different flavors for the backlight. The most feature-rich one has three additional WS8211 RGB leds (two status leds and the last one for the LCD). Thus, additional code probably needs to be included to enable the LCD backlight. A simple example for a custom event can be found [here](https://gist.github.com/CyberDeck/47ecf2b39a128975c38f36750cc56a51).

Additionally, this pull request fixes a yet unknown inconsistency introduced by commit 55970d37633edc06f0224f43fa494489f842c7fa. The `ui.cpp` files are not consistend between the DUE and AVR directories. The code from DUE made more sense to me, so I copied over the changes to the AVR directory. By doing so an filament change will first wait until end of all moves and than pause the extruders (not vice versa). Please double check that my pull request will not introduce an additional bug.


repetier:
Thanks for the driver settings. Have also included it now in the config tool.

PR #877, opened 2019-06-15
Increase MAX_ANALOG_INPUTS to 14 for Due.

dominikacri:
SAM3X has 14 and SmartRamps use A11,A12,A13 as TEMP_PIN which otherwise are not addressable.


repetier:
You are right. Even 16 pins are possible for due chip, so upgraded it to 16 channels.

PR #879, opened 2019-07-07
Use tempError / temp difference in calculating decoupling variance

dominikacri:


repetier:
Good seen error. In line 147 the report still reports error not tempError but have changed this for next update. But that needs to wait until my mods for tmc are included (thanks for driver packe).

PR #882, opened 2019-07-21
- Fix wrong position after homing Z with z Probe and zoffset for xy

dominikacri:
After first G28 position (real vs. corrdinates) was wrong. Second time worked because then all axes were alligned. 
Unsure this is the right fix for all ZProbe type cases but it fixes the bug for zprobe 1 with offsets.
Fix is in setToolOffset

### DIFF
--- a/src/Repetier/src/communication/GCodes.cpp
+++ b/src/Repetier/src/communication/GCodes.cpp
@@ -237,6 +237,7 @@ void GCode_30(GCode* com) {
         Motion1::moveByOfficial(Motion1::tmpPosition, Motion1::moveFeedrate[Z_AXIS], false);
         float zheight = ZProbeHandler::runProbe();
         if (zheight == ILLEGAL_Z_PROBE) {
+             GCode::fatalError(PSTR("G30 probing failed!"));
             return;
         }
         float zProbeHeight = ZProbeHandler::getZProbeHeight() + startHeight - zheight;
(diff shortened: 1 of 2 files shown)


repetier:
The small fixes are ok and I have taken them in my next update. The setToolSetOffset is not correct I think. It includes a z move so it should have z homed that way. But I changed it for next update to

```
void Motion1::setToolOffset(float ox, float oy, float oz) {
    if (Motion1::isAxisHomed(X_AXIS) && Motion1::isAxisHomed(Y_AXIS)) {
        setTmpPositionXYZ(currentPosition[X_AXIS] + ox - toolOffset[X_AXIS],
                          currentPosition[Y_AXIS] + oy - toolOffset[Y_AXIS],
                          currentPosition[Z_AXIS]);
        moveByOfficial(tmpPosition, Motion1::moveFeedrate[X_AXIS], false);
        waitForEndOfMoves();
    } else {
        currentPosition[X_AXIS] += ox - toolOffset[X_AXIS];
        currentPosition[Y_AXIS] += oy - toolOffset[Y_AXIS];
        updatePositionsFromCurrent();
    }
    toolOffset[X_AXIS] = ox;
    toolOffset[Y_AXIS] = oy;
    updatePositionsFromCurrentTransformed();
    if (Motion1::isAxisHomed(Z_AXIS)) {
        setTmpPositionXYZ(currentPosition[X_AXIS],
                          currentPosition[Y_AXIS],
                          currentPosition[Z_AXIS] + oz - toolOffset[Z_AXIS]);
        moveByOfficial(tmpPosition, Motion1::moveFeedrate[Z_AXIS], false);
        waitForEndOfMoves();
    } else {
        currentPosition[Z_AXIS] += oz - toolOffset[Z_AXIS];
        updatePositionsFromCurrent();
    }
    toolOffset[Z_AXIS] = oz;
    updatePositionsFromCurrentTransformed();
}
```

So we apply first xy offset if possible and then Z offset. That way after xy homing z probe xy offset can be applied physically which I think is the main point for the correction.

PR #891, opened 2019-11-26
Support for TMC2209 drivers.

AbsoluteCatalyst:
I'm running the newest TMC2209 drivers that were recently released on my machine. I added some support for them with repetier V2 on a small forked repo of mine a little while ago. It's just a copy of the current TMC2208 classes/setup with some minor changes (individual UART addresses, stallgaurd4 etc).

- Stallguard4 uses values between 0-255. The stall sensitivity EEPROM/respective values were replaced with short instead of byte to support the new range (and the old range).

- I quickly added on a new flag (32) on line 132 for stepper.cpp just to display the correct value range to the user if they're using stallguard4 vs stallguard2. Could possibly be handled differently.

- I'm running these with the baudrate set to 500000, though they should support up to 700000 baud (stock fclk)

- Haven't added any coolStep configuration options for them. I don't believe the options are considerably important for now, and I noticed some other driver definitions currently lack them as well. 

I was originally planning to add in some pin interrupt handling for the diag output on these at some point, but I realized repetier wrote some support for this already. (ENDSTOP_STEPPER ?) 

([b774dea](https://github.com/AbsoluteCatalyst/Repetier-Firmware-V2/commit/b774dea387bfa8a53ba3013063f2d52fb5265e14) original commit w/some info.)

### DIFF
--- a/src/Repetier/src/drivers/stepper.cpp
+++ b/src/Repetier/src/drivers/stepper.cpp
@@ -96,7 +96,7 @@ void AdjustResolutionStepperDriver<driver>::eepromHandle() {
 
 #define PSB_STALL_SENSITIVITY_POS PSB_STEALTH_POS + PSB_STEALTH_SIZE
 #if STORE_MOTOR_STALL_SENSITIVITY
-#define PSB_STALL_SENSITIVITY_SIZE 1
+#define PSB_STALL_SENSITIVITY_SIZE 2
 #else
 #define PSB_STALL_SENSITIVITY_SIZE 0
 #endif
@@ -129,8 +129,10 @@ void ProgrammableStepperBase::processEEPROM(uint8_t flags) {
     }
 #endif
 #if STORE_MOTOR_STALL_SENSITIVITY
+    //Note, flag 32 is stallguard 4, we don't use that flag for anything else yet, other than a small change in eeprom text.
     if (flags & 16 && stallguardSensitivity != -128) {
-        EEPROM::handleByte(eprStart + PSB_STALL_SENSITIVITY_POS, PSTR("Stall Sensitivity [-64..63]"), stallguardSensitivity);
+        EEPROM::handleInt(eprStart + PSB_STALL_SENSITIVITY_POS, (flags & 32) ? 
+                PSTR("Stall Sensitivity [0..255]") : PSTR("Stall Sensitivity [-64..63]"), stallguardSensitivity);
     }
 #endif
 }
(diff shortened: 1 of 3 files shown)


repetier:
Thanks for the patch. Did not have the TMC2209 for testing which is why I omitted it. Nice to see that the structures are understandable enough that others can extend them.
I have already merged it and updated the documentation to include them.
Using flag 32 for the message is ok. Good way to signal the difference without any complicated code handling.

Regarding end stop forget ENDSTOP_STEPPER. That was before I learned how it works. The diag pin just switches signal for a short time to indicate the stall. So just use ENDSTOP_SWITCH_HW with a pin connected with diag and use that as end stop/ z probe. There is no other handling required if stallguard is enabled.

AbsoluteCatalyst:
Awesome. That was quick.

As for the diag pin; thanks, I was wondering how to properly configure it at the time. 

But one thing I still think about: Their diagnostic outputs can also trigger in short conditions, and besides homing, how about having stall detection enabled during printing?

My original concept was to have an interrupt input solely for stepper driver diagnostic events in events.h where-in users can easily add in their own ways of handling events such as shorts and stalls at any point. Most likely being immediately lower the bed -> home -> then return to the print and continue. 

Though I'm sure this may interfere with the movement planner, as well as potentially not returning to the precise position it once was. I feel as if I once read a conversation between you and a user explaining this on the repetier forums but I can't find the post.

Regardless; thank you for your work on V2, the new movement planner has definitely improved my print quality.

P.S. I have a small set of additions I've created that I was thinking of possibly creating pull requests for. (I'd have to separate some readme file changes first, among other things.), though I'm unsure if you have any requirements/limitations on the kinds of additions and changes you're willing to merge onto the dev2 branch? Nor do I know if you have any interest in such at the moment as some aren't necessarily core functionality.

repetier:
Are you sure that shorts also trigger diag1? I mean you can configure it so it is not a must I guess. Anyhow I see no real problem in it. A short would ruin the print anyway.
Mapping event and then read status is a bad idea I fear. Any complex communication should only be done in main thread. So capturing it with a hardware signal is at least out of order. So you need to do it software wise with a frequent pull, but that will add a delay since you can not do it in stepper interrupt for same reason also it would need to be there. And with a delay you can not relyable home any more to identical positions.

Regarding additions - the idea of the module system was to make it easier for others to add specilized solutions for different hardwares or problems. And since I do not own all hardware to test or develop I was hoping to get additional parts from somewhere. That is why it is open source after all. And since only parts being used get linked it is not a drawback to have more modules available than being used.
So yes, if you have additions that are written as general modules (so no assumptions about pin numbers etc, so users can configure it) they are welcome.

AbsoluteCatalyst:
Ah, after quickly glossing over some of TMC's datasheets it appears the shorted diagnostic output is just specifically a feature for the 22xx series. I didn't realize this. Latches the diagnostic output until reset via ENN, as well as setting s2vs/s2g in the DRV_STATUS register. I haven't personally tested it myself, though I'm going to experiment with it soon out of curiosity. I see your point with the driver communication. 

And alright, I did design the additions to be extensions to the new general module system/user configurable. Good to know.

PR #892, opened 2019-12-03
Allow ARM-based (due) boards to attach endstop hw switches to any pins.

AbsoluteCatalyst:
Small fix to bypass digitalPinToInterrupt checks for arm boards which may have pins defined beyond the 66 that function limits to.


repetier:
Nice addition for the time when AVR is working. Thanks.

PR #893, opened 2019-12-03
Add support for Due Timer PWM's.

AbsoluteCatalyst:
Set up timer PWMs for SAM3X8E. Allows all the possible timer pins to output PWM, including TIOA* TIOB*. 

Original Commit Desc:
Seamlessly integrated hardware based timer PWM mode for all the timer pins supported on SAM3X8E based boards. It's done alongside the normal hardware PWM io configuration. The only problem is you're currently still allowed to overlap/overwrite the original timers used for motion/beeper/servos, be warned as to which pins you use and which timers they belong to, to avoid this issue.

I have some code i'm currently working on which dynamically reassigns these stock timers to use any other available timers not used by the user, since we want the user's channels and configuration to have a bit of a higher priority.

### DIFF
--- a/src/Repetier/src/boards/due/HAL.cpp
+++ b/src/Repetier/src/boards/due/HAL.cpp
@@ -147,6 +147,79 @@ void HAL::setupTimer() {
 #endif
 }
 
+struct TimerPWMPin {
+    TimerPWMPin(int _pin, Pio* _pio, uint32_t _pio_pin, byte _tc_channel, bool _peripheral_A) : 
+        pin(_pin), pio(_pio), pio_pin(_pio_pin), tc_global_chan(_tc_channel >> 1), tc_local_chan(_tc_channel / 6), peripheral_A(_peripheral_A), tio_line_AB(_tc_channel & 1) { 
+            if (tc_local_chan == 1) {
+                tc_base_address = TC1;
+            }
+            else if (tc_local_chan == 2) {
+                tc_base_address = TC2;
+            }
+            else {
+                tc_base_address = TC0;
+            } 
+        }
+    int pin;
+    Pio* pio;
+    uint32_t pio_pin;
+    byte tc_global_chan; // 0 .. 8 What's our overall timer channel number?
+    byte tc_local_chan; // 0 .. 3 Which timer are we in? 
+    bool peripheral_A; // Do we need to set the peripheral to A instead of B?
+    bool tio_line_AB; // 0 = A, 1 = B Is this the TIOA or TIOB output pin?
+    Tc* tc_base_address; // TC0 .. TC2 registers
+};
+
+// Each timer COUNTER has 3 timer channels.
+// Each timer CHANNEL has a TIOA* and TIOB* pin where * is the number of the timer channel.
+#define NUM_POSSIBLE_TIMER_PINS 18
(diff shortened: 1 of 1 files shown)


repetier:
Reassigning timers sounds interesting.
BTW: Currently the timers are selected so each has it's own prescaler. Now with setting pwm we allow all frequencies. Do we get into troubles if these differ from the one used for our Stepper/PWM/Servo timers?

AbsoluteCatalyst:
Sorry for the late reply!
As long as we use a different timer channel number than the stepper/servo timers, there isn't any issue between them and the PWM timers. The prescaler/clock selected is a channel specific setting, so no issues.

If we *do* use a pwm-timer pin that's on one of the channels used for the stepper/servos, the user's pwm-timer just has it's mode settings (frequency etc) overwritten by the stepper/servos timers that're initialized later. _(My original commit's description had this the wrong way around.)_

a quick example: 
My air pump (fan 0) is on Pin 2 which ends up being TC0_CHA0 (Timer channel 0, line A) and I have it set to 250Hz. 
This conflicts with the MOTION2_TIMER, and so after setup my air pump will actually be running at PREPARE_FREQUENCY (2kHz) since my pump's pwm-timer channel was overridden by the init of the motion2 timer. 
The pump still functions anyway too, since changing it's PWM value adds on the RA/RB compare flags to whatever the pump's pwm-timer channel is (In this case the motion2), and also scales off of whatever the channel's current RC compare value is. 

(To get around this overwrite issue I just moved the motion2_timer to channel 2 (ID_TC2) on my machine.)

repetier:
Ok, I think I get it. Please correct me if I'm wrong. Frequency depends on RC value being the value where timer interrupt is called. This exists for 3 timers with 3 channels so in total 9 times. PWM is set with A or B register so we get at maximum 9 * 2 = 18 PWM outputs where 4 are bound with frequency.

The regular timer interrupts are set after PWM so they overwrite frequency to what they need. You read set RC value so set pwm is always correct even if it changed.

I use 5 timers internally where frequency is set so pins on that are fixed:
Motion 2 timer - pin 2 + 13 - PREPARE_FREQUENCY (500 - 2000)
PWM Timer - Pin 61 + 60 - PWM_CLOCK_FREQ (10000)
Motion 3 Timer - Pin 11 + 12 - STEPPER_FREQUENCY very high
Servo Timer - Pin 5 + 4 - SERVO_CLOCK_FREQ (1000 hz, if servos are present)
Beeper Timer - Pin 108 + 109 - Varying frequency if BEEPER_PIN > -1, 
Beeper also sets TC_SetRA in tone, also I think that is not necessary. Anyhow with varying frequency it is bad to use that timer anyway.

If this is correct I will add it into the docs. Otherwise users will never understand what is allowed and not.

AbsoluteCatalyst:
Yeah that's basically it, the only last thing is that while we can pwm both TIOA and TIOB pins in a single timer channel with independent duty cycles (TC_SetRA and TC_SetRB)
they're both restricted to the same frequency set for their channel (TC_SetRC) 

So pins 60/61 would have the same frequency if they're both used, etc.

As for *who's* frequency do they both use? The frequency of the last pin initialized for that channel. I only state this for pins on timer channels who aren't overridden by the internal timers.

repetier:
Thanks. I have now updated the docs accordingly.

PR #894, opened 2019-12-03
Output EEPROM at any point using emergency parser.

AbsoluteCatalyst:
Added M205 to the emergency parser as well to let repetier-server's interface read the eeprom even during long wait commands. 

In the original commit I mentioned that it only works around 7 times during waits. 
I believe I fixed this by adding M205 to the <PriorityCommands> tree on repetier-server's firmware xml file.


repetier:
Not exactly an emergency to run it during heatup, but also had the wish to run it then. Mainly to check if I changed that one value before starting print. Only drawback is that it blocks M108/M112 if being run so I added it. Will also add it in server to emergency commands to make sense.

PR #895, opened 2019-12-05
Add M106 D<delay> parameter for fans.

AbsoluteCatalyst:
Think this small fan delay/timeout functionality is useful to implement since more and more people are using enclosures that have fan/exhaust systems or more exotic configurations.

I chose the D parameter to avoid conflicting with any other firmware's gcode parameters (based off the [reprap g-code wiki](https://reprap.org/wiki/G-code#M106:_Fan_On))

Also added a EVENT_FAN_TIMEOUT(fanId, targetSpeed) event. Good for custom machines with custom actions (eg. motorized vents.)

Fan timeouts are checked with the 500ms timer inside Commands::checkForPeriodicalActions

I removed setFanSpeedDirectly as it seemed redundant and didn't see it in use anywhere else? I assumed it was a carry-over from V1. My apologies If you had plans for it.

<br/>
<i>(Updated)</i> Original Commit Desc: 

Adding a D\<seconds> parameter to the M106 command delays the adjustment of a fan's speed.

Example:
At the end of a print, run:
```
M106 P1 S255
M106 P1 S0 D180
```

^ The first M106 turns on fan P1 (enclosure fan, etc) at max.
The following M106 queues P1 to shut off after 180 seconds (3 minutes.)

You can queue any kind of S<speed> parameters to occur in the future using D\<seconds>

Any new M106 commands will overwrite the previously queued timeout. If no D parameter is included, the delay/timeout is simply cleared If there is one.

I've been using this for my printer's enclosure fans. At the end of a print my fans stay on for a short while to cool off the print. 

Sidenote: Removed duplicate setFanSpeed found in commands.cpp as well as setFanSpeedDirectly.

### DIFF
--- a/src/Repetier/src/Events.h
+++ b/src/Repetier/src/Events.h
@@ -158,4 +158,8 @@ Each of the following events describe the parameter and when it is called.
 #define EVENT_CUSTOM_TOOL_CHANGE_SETUP(tool) \
     {}
 
+// Called when a fan reaches it's M106 D<seconds> timeout
+#define EVENT_FAN_TIMEOUT(fanId, targetSpeed) \
+    {}
+ 
 #endif // EVENTS_H_INCLUDED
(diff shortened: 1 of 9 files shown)


repetier:
At first I had no idea where to use it, but the setting at end of print really makes sense. Much better than adding G4 delay for same purpose.

I have renamed the struct to start with capital letter (used for all classes, structs) and to accept ms instead of seconds. Not that it makes much different since it is in 500ms timer, but that makes it more future proof I think.

PR #896, opened 2019-12-09
Due Overridden TIOB TC_CMR_EEVT_XC0 Fix

AbsoluteCatalyst:
If our TIOB was overridden by an internal timer this small bit needs to be reapplied again to keep TIOB set as an output. 
Forgot about that small quirk on the SAM3X8E.


repetier:
Thanks for the update. Would be hard to find that solution if it happened.

PR #897, opened 2019-12-13
EEPROM PID update derived ki/kd

AbsoluteCatalyst:
eeprom wasn't updating derived ki and kd values after being saved manually.


repetier:
Great. Thanks.

PR #898, opened 2019-12-16
Tone buffer

AbsoluteCatalyst:
A small tone queue buffer/processor for M300 commands. Allows the beeper to function concurrently to other functions. Currently uses the Commands::checkForPeriodicalActions function to process them.

Added an EEPROM setting for tones/sounds, as well as a menu option to toggle them. 

Function to add a tone internally is just: 
Printer::addToToneQueue({frequency, delay}); 
Adding a tone without any frequency will just create a delay between tones. 

I had an idea to create something similar to the io_light module for tone themes for UI actions such as clicks, confirmations, errors, etc.


repetier:
Nice update. When I yesterday fixed broken tone on v1 due I thought that it is not so good to block main thread with tones as temp. updates etc. get delayed, so nice to have the problem already fixed:-)

With light module I see what you mean, but I think it is not so easy as we now also have animations. Actually we have now a solution to show the current state. That was planned also to show preheating, error state etc. 
What I think would work is we had a switch in the code - queue empty = current solution, queue filled with timed settings - we show the queue for the light source. And once it is empty we go back. 
The main problem is also you normally have only one light source that is not as it must be. We can have any number of light source - one case light, one fo temperature, ... so with animations each light source needs an animation buffer.

AbsoluteCatalyst:
It would be a bit complicated to set up. I was thinking running alongside it/in parallel as a separate module, though perhaps it may be better suited using the events.h system for it and only creating tone tables in io? My original concept looked something like this:

BEEPER_TONE(<source>, <cond>, <size of tone table>, <tone table>, <repeats>) 

so: 

IO_PWM_HARDWARE(Beeper1PWM, BEEPER_PIN, 0)
BEEPER_SOURCE(Beeper1, Beeper1PWM)

BEEPER_TONE(Beeper1, (nextAction == GUIAction::CLICK), 1, {14500, 10}, 0) //UI Button Click
BEEPER_TONE(Beeper1, (HAL::eprSyncTime == 0), 2, {{19500, 30}, {15500, 50}}, 0) //EEPROM Synced
(The eeprom synced tones is a weird personal example of something I have setup right now.)

But I realized doing it via conditionals as above may be a bit weird to implement vs using the events. 


I quite like the lights module, and your post just gave me an excellent idea I didn't think of for some red preheating case lights inside my enclosure I want to setup now.

repetier:
Why adding a pwm if we already have your tone play list? Wouldn't it be easier to just add the sounds to tone play list?

For BEEPER_TONE you should add a own name so you can generate a static variable with the melody stored in flash (for avr) and read it from there. Using eeprom does not sound nice here.

One thing is where to add the macros. If you have a bool to store if it last matched you could do it in periodical actions. I mean it should first be false before you play it again. In periodical actions or 100ms timer it would have access to all variables and not be just limited to e.g. new nextAction.

AbsoluteCatalyst:
Good point about the pwm/source bit, it wouldn't be necessary to add that parameter to the tone list macro. 
I think in either of those timers it would make sense, and the repeats parameter would coincide well with that. 
I'll write up a quick module like this soon.

PR #900, opened 2019-12-18
Due TimerPWMPin tc_local_chan fix

AbsoluteCatalyst:
tc_local_chan wasn't being calculated correctly and didn't allow timers to be set on their proper respective channels. Somehow I didn't catch this issue before, or perhaps I fudged it during a diff in meld. Apologies. I've double checked as many timer pins as I could and all seem to be in working order again.
- Struct readability. 
- Clang formatted.


repetier:
Thanks.

PR #901, opened 2020-01-02
Implements M73 prgress command

ondiiik:
Some slicers provides very good **time estimation and progress and exports it into GCODE** in the form of `M73` command.

This command provides progress in percentage as rgument `P` and remaining print time in minutes as argument `R`.

As there can be many types of displays, as most easy way seems to be to use message on display (same as command `M117`). Message with progress is displayed always when `M73` code is received and looks like in following example:

`[...74.   ] Rm 1:36`

Just note that `UIDisplay::setProgress` is here in the role of interface method, which may be modified in future to controll real progress bar on displays which supports that.

### DIFF
--- a/src/ArduinoAVR/Repetier/Commands.cpp
+++ b/src/ArduinoAVR/Repetier/Commands.cpp
@@ -2268,6 +2268,10 @@ void Commands::processMCode(GCode* com) {
             }
         }
         break;
+    case 73: // M73 - Print status on display
+        UI_PROGRESS_UPD(com->hasP() ? com->P : -1,
+                        com->hasR() ? com->R : -1);
+        break;
     case 80: // M80 - ATX Power On
 #if PS_ON_PIN > -1
         Commands::waitUntilEndOfAllMoves();
--- a/src/ArduinoAVR/Repetier/DisplayList.h
+++ b/src/ArduinoAVR/Repetier/DisplayList.h
@@ -1963,6 +1963,7 @@ inline void uiCheckSlowKeys(uint16_t &action) {}
 #define UI_STATUS_UPD_F(status) {uid.setStatusP(status);uid.refreshPage();}
 #define UI_STATUS_RAM(status) uid.setStatus(status);
 #define UI_STATUS_UPD_RAM(status) {uid.setStatus(status);uid.refreshPage();}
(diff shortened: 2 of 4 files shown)


repetier:
Normally I'd say nice done, but firmware already supports the progress capability to show layer/filename/percentage to user. So this is an alternative solutions that contradicts with our host/server time prediction being send which can lead to confusion if users get 2 different times.

Progress send by progress capability:
M532 X{f:progress} L{i:currentLayer}
Also you need M530-M531 to use it.

In sd print mode this also gets activated automatically but would not hurt since then time is not shown as that comes from host.

I think I take it with the modification to be ignored when isPrinting() is true and sd printing is not active. Then it is an alternative in case host does not activate progress display and it is a win in that case.

ondiiik:
Ok ... it sounds that at least progress part is like reinventing a wheel ;-). Then for the progress there is really better way to use `M532`, _(or rather say redirect percentage from `M73` to call functionality of `M532` from inside)_.

The question is _remaining time estimation_, which is calculated quite precisely in slicer and is also stored inside. Is this also solved somehow by some other `M` code?

Anyway, I would expect that slicer will work either in _Repetier Host_ mode producing `M53x` code with progress and layer number or in `M73` mode _(in my case generated by PrusaSlicer in Repetier mode, but with M73 enabled)_ producing `M73` instead. From this point of view it seems to me that both code may not appear together in one single G-file, so both may be supported by firmware. What is your opinion about this?

Anyway, I'll modify at least progress to be calling native repetier progress (`Printer::progress`) instead of printing message. The rest I would have to think little more about it ;-)

repetier:
No need to change anything I think. I will add the one line that is missing.

You missunderstand how hosts work and the problem with M73 - so far I only know slic3r computes it but precision depends on firmware and it's settings. Therefore e.g. repetier-server also simulates a complete print including jerk and acceleration and cooling/heating time making it very precise if configured correctly. Main advantage is that it is slicer independent and also works with slicers not adding M73. And hosts show their internal time estimate, so in that case M73 has to be ignored. So your solution should be used if hosts do not use their own solution e.g. not enabling progress show mode. So what I want to do is just add a test at your M73 if we have a case where it should be ignored or not. That way it is a useful addition.

ondiiik:
Ok ... now I see ;-)

PR #902, opened 2020-01-17
Radds smartui

furynick:
Adds pin configuration for RepRapDiscount Smart Controller for RADDS

### DIFF
--- a/src/ArduinoDUE/Repetier/DisplayList.h
+++ b/src/ArduinoDUE/Repetier/DisplayList.h
@@ -319,6 +319,29 @@ void uiCheckSlowKeys(uint16_t &action) {}
 #define UI_RESET_PIN -1
 #define UI_INVERT_MENU_DIRECTION 1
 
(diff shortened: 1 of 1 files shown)


repetier:
How did you connect the RADDS with controller? As far as I know there is no standard connector matching or is there now an official adaptor available?

furynick:
Wire by wire, there are some adaptors out there but no official one.

I posted a photo on repetier forum, I guess the previous comment (thingiverse) is the best way to connect it.

https://forum.repetier.com/discussion/4876/repetier-1-0-0-due-radds-1-5-reprapdiscount-full-graphic-smart-controller-128x64#Comment_30091

repetier:
Just saw that this is agains the wrong branch. Should be dev to be included. But checking dev version I see we have it already added for this adapter
https://www.thingiverse.com/thing:1740725/files

but pins are the same just some more:-)
So I close this for being already included.

furynick:
Good news, thanks.

PR #911, opened 2020-03-07
add LIGHT_SOURCE_PWM

AbsoluteCatalyst:
Small add-on for the io_lights module. PWM smoothed/fading light sources as well as M355 P<0..255> for general case light dimming.
Works with hardware and software PWM sources. 
The new macros are:
LIGHT_STATE_PWM(PWMCaseLightState)
LIGHT_SOURCE_PWM(caseLight, PWMCaseLight, PWMCaseLightState)
LIGHT_COND has a new 7th parameter for brightness (0-255) 
eg.
`LIGHT_COND(PWMCaseLightState, true, Printer::caseLightMode, 255, 255, 255, Printer::caseLightBrightness) `

The light fading takes place in the periodical actions timer and the default refresh rate is 30 milliseconds. Can be easily changed in io_light.h inside the LightStorePWM class. (beyond 50ms it becomes fairly choppy though.)

The duration of each fade transition is hard-coded at the moment in io_light.cpp, under the LightStorePWM::set method. (computePWMStep(durationMS, brightness);) 

computePWMStep isn't such a complicated function, though it's more-so future-proofing in case it's changed around. It's only called on brightness changes or mode changes.

 I added a small 8-bit int rollover constraint function for the pwm stepping to support AVR platforms later in the future.

### DIFF
--- a/src/Repetier/src/PrinterTypes/Printer.cpp
+++ b/src/Repetier/src/PrinterTypes/Printer.cpp
@@ -85,6 +85,7 @@ bool Printer::failedMode = false;
 #define CASE_LIGHT_DEFAULT_ON 0
 #endif
 fast8_t Printer::caseLightMode = CASE_LIGHT_DEFAULT_ON;
+fast8_t Printer::caseLightBrightness = 255;
 
 #if defined(BEEPER_PIN) && BEEPER_PIN >= 0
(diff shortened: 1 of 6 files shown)


repetier:
I really like this. Shows flexibility of module system and adds a cool feature. Will document it soon when I'm through with the probing bugs.

PR #912, opened 2020-03-07
Due fix low frequency pwm

AbsoluteCatalyst:
this was slightly broken and was giving us the clock division rather than the clock divisor index. now works down to ~1.25hz on the due if necessary


repetier:
See my error here. Just wonder why I didn't show up. Would expect it to make more problems when using it for temperature.

AbsoluteCatalyst:
It functioned properly as long as the frequency was higher than 320Hz I think, though it was pretty elusive. Never noticed it either until I was experimenting with the lights/servo pwm's.

PR #913, opened 2020-03-07
CoolerManager max pwm fix

AbsoluteCatalyst:
max pwm limit was occasionally being bypassed. small typo i think


repetier:
Yes, think you are right here:-)

PR #919, opened 2020-04-01
Fix PWM light condition execution conflict

AbsoluteCatalyst:
I missed this (and the previous statics, thanks)
Last winning condition now sets the mode and brightness properly on the next reset


repetier:
Sounds good to delay it to next cycle. Thanks.

PR #920, opened 2020-04-01
IO_PWM_RAMP add-on

AbsoluteCatalyst:
Soft starting/soft stopping on pwm outputs. Helps reduce some of the violent surge/inrush current spikes when turning on high current loads (eg air cooling pumps, motors). 
Start delay/end delay are set independently. The threshold parameter that dictates how far away the target PWM must be from the current PWM in order to ramp up/down.

Ramping is done in the 100ms timer just like kickstart.
IO_PWM_RAMP(Pump1PWM, Pump1PWMHW, 5, 4, 45)
// 500ms ramp-up, 400ms ramp-down, threshold of 45.

It can be used on heater outputs too, but you may need to recalibrate your PID/dead-time settings.

### DIFF
--- a/src/Repetier/src/io/io_pwm.h
+++ b/src/Repetier/src/io/io_pwm.h
@@ -41,6 +41,7 @@
 
 #undef IO_PWM_SOFTWARE
 #undef IO_PWM_KICKSTART
+#undef IO_PWM_RAMP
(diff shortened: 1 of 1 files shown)


repetier:
Cool idea. Will update docs for the function.

PR #921, opened 2020-04-02
20X4 display bed/chamber temp cycle

AbsoluteCatalyst:
also blinks ???.? if z isn't homed.


repetier:
Great idea with the switching bed/heater. Maybe I can make the graphic version better with that as well:-)

PR #922, opened 2020-04-07
IO_TEMPERATURE_BETA

AbsoluteCatalyst:
Option for Steinhart-Hart/Beta generated thermistor readings. The C Coefficient is optional and can just be set 0 for a basic beta model instead. 

example:
IO_TEMPERATURE_BETA(TempExt2, IOAnalogExt2, 3950.0f, 5100.0f, 10000.0f, 0.0f)
IO_TEMPERATURE_BETA(name, analog, beta, seriesResistance, thermistorR25, cCoefficient)

### DIFF
--- a/src/Repetier/src/io/io_temperature.h
+++ b/src/Repetier/src/io/io_temperature.h
@@ -42,6 +42,7 @@ IO_TEMPERATURE_TABLE(name,analog,table)
 #undef IO_TEMP_TABLE_NTC
 #undef IO_TEMP_TABLE_PTC
 #undef IO_TEMPERATURE_TABLE
+#undef IO_TEMPERATURE_BETA
 #undef IO_TEMPERATURE_MAX31855
 #undef IO_TEMPERATURE_MAX6675
 #undef IO_TEMPERATURE_FAKE
@@ -53,6 +54,7 @@ IO_TEMPERATURE_TABLE(name,analog,table)
 #define IO_TEMP_TABLE_NTC(name, dataname)
 #define IO_TEMP_TABLE_PTC(name, dataname)
 #define IO_TEMPERATURE_TABLE(name, analog, table)
+#define IO_TEMPERATURE_BETA(name, analog, beta, seriesResistance, thermistorR25, cCoefficient)
 #define IO_TEMPERATURE_MAX31855(name, spiDriver)
 #define IO_TEMPERATURE_MAX6675(name, spiDriver)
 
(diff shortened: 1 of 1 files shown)


repetier:
Looks good and had totally forgotten that I did not have that one. Have replaced const with constexpr so compile computes the constants at compile time. With pure const compiler can decide to do it compile time or run time.

AbsoluteCatalyst:
Ah whoops, that's what I was aiming for. Forgot const alone wasn't strict. Thanks.

Speaking of which I was meaning to ask something related, 
The NUM_* macros for the braced init lists within configuration.h ; why not make them constexpr sizes within the io_* class defs to avoid setting them manually? Or is the current overall setup preventing that atm/design choice?

repetier:
How would you do that? I know with
std::extent<decltype(MyArray)>::value;
I can get array size of an array, but the arrays are initialized in cpp files and I need the size also in other cpp files. Or did I miss a solution that allows it in all cpp files and maybe even some .h files (NUM_TOOLS is used there as well)? I really would like to remove this source of errors.

Speaking of errors I managed to modify the code a bit so that I could at least check if set values are correctly like this:
```
    constexpr int numTools = std::extent<decltype(Tool::tools)>::value;
    static_assert(numTools == NUM_TOOLS, "NUM_TOOLS not defined correctly");
```
Will add that for all arrays to be at least sure the sizes are set correctly.

Alternatively the sizes could be set in a variable and read from there, which is the only other solution I see. Unfortunately it is not allowed to define extern constxpr :-(

AbsoluteCatalyst:
I remembered I wanted to make a io_beeper module a while ago and so was working on it now. 
The way I originally set up the number of beepers is through an extern variable within io_beeper.h in the class definition, and then a constexpr define under the define variables.
But errr, now that I double checked; I'm wrong and i'm _not_ creating it during compile-time this way as I thought.
```
#elif IO_TARGET == IO_TARGET_CLASS_DEFINITION

extern BeeperSourceBase* beepers[];
extern ufast8_t numBeepers;

#elif IO_TARGET == IO_TARGET_DEFINE_VARIABLES

BeeperSourceBase* beepers[] = BEEPER_LIST;
constexpr ufast8_t numBeepers = (sizeof(beepers) / sizeof(beepers[0]));
```
(so in this case constexpr is just ignored.)

However, I _am_ able to use numBeepers effectively in any other .cpp file without issue, and most .h files (depending on include order etc) Same goes with the beepers brace list, while still avoiding a macro define. Perhaps not exactly what we're looking for though

repetier:
constexpr ufast8_t numBeepers = (sizeof(beepers) / sizeof(beepers[0]));

should be

const ufast8_t numBeepers = (sizeof(beepers) / sizeof(beepers[0]));

and you can use it. It is const but not compile time. Only thing is it gets a variable defined with a value evaluated at compile time. Just what I said - I could put length in a const variable. Only thing is that does not allow many compile time tests like I do e.g. for NUM_AXES that omit A_AXIS etc if lower on optimized versions. 

I see a few places where it might work - e.g. temperature tables we can use it as it is only used there. Will at least test that - don't like counting all the entries. For the rest I will add static_assert so we are on safe side at least.

PR #929, opened 2020-05-08
Import/Export bump matrix to CSV. M323 S2 Scan & Auto import

AbsoluteCatalyst:
This is a bit of a work-in progress, but this let's us import/export (M375 & M374) the bump matrix to a human readable/editable CSV file on the SD card, as well as an addition to M323 that'll scan for bump matrix files on the SD and auto import based on current target temperature.

image

(I designed the layout to be vaguely similar RepRapFirmware's exported mesh csv's as I found them easy to read, but they're not interchangeable.)

For importing CSV's I wrote a small header-only parsing library (CSVParser.h) that helps navigate fields and cells. I added documentation to the functions available. 
For exporting it's using the SDFat Library's basic printFields/writes which is a tad bit messier than desired at the moment.

Command to export: M374 filename. Importing: M375 filename. 
(.csv is appended if it's not already included.)

G32 now saves the target temperature the bed was probed at and includes it in the EEPROM and exports. (**Note**! EEPROM will shift a bit) 

M323 "S2" scans the sd card root directory for any CSV files that're valid bump matrix files and finds one closest to the current target temperature of the bed and grid size, and imports them. 
D<difference> sets the maximum temp difference between the file and current target temp. By default it's 10c. More and the file isn't considered. 

My current concerns/issues/todo right now:
- Would it be better to import/export the transformation matrix itself or the plane then regenerate the transformation? back-tracking the source, I can't see any conflicts using the matrix itself I think.
- I'm not entirely sure If adding the scanning to M323 command is suitable, perhaps G33? 
- Might be better to move the scanning process to it's own method. 
- Improving CSV exporting a bit (proper write to cells/columns). This way it could be used later for other features (eg. logging/csv module/etc)
- Have the CSV files in their own sub-directory on the SDCard? 

I hope I haven't missed anything.


repetier:
Ok just checked the sources. So my comments to what I saw and your questions:
- I think grid temperature should also be reported in eeprom so you can see where it belonged to.
- Consider that several leveling methods exist but only one has grid so secure the M374/M375 for compile errors on the other variants.
- The grid it self contains deviations against the currently used autolevel matrix. So with any other matrix the result would be wrong. And since you expect different heights/bumps G32 would compute a different average plain from the heights so using it with another one leads to different results. So I think you need to store the 9 values of autolevel as well and restore them. To be exact you also need to add z length/height as well especially for printers using z max homing!
- M323 for scanning is ok I think. It is like enabling just with best fit for a given temperature.
- Own sub directory would be better. If you have many you do not want them in root level amoung the gcodes. Also it gets clear that the csv files in that folder are for grid levels, so we can reuse extension at other places if required.
- If you have an own directory would a naming scheme not easier to find the right file. E.g. if they are just named grid_<temp>.csv you only need to scan directory and load best. Or instead of grid_ the name so you can have several groups in case you have different beds/metal sheets like Prusa. So you still have the temperature for quick scan.
- There might be a clear folder command needed as well in case bed changed position or something else changed.
- CSVParser.h should have a copyright block with a hopefully compatible license. Is it from you or a deviation of an existing one? I also saw it is not compatible with avr processors. If you use PSTR with avr it is stored in flash and need to be read from there. So the solution only works with arm 32 bit processors. 

Since I also heard that temperature is a big factor for some printers I like the implementation. No need to use but possible if required and easy to handle with sd card.

AbsoluteCatalyst:
I don't get so so many drastic changes myself since I'm running a glass bed, though it at higher temperatures (120c+) it does seem to produce a more noticeable "pillow" affect. I too remember some people stating they did experience large changes though. I imagine metal bed printers may experience more.

* Do you mean as a non-silent handle EEPROM value? ATM there's a print for the bed temp appended to reportDistortionStatus() so G33 L reports it
* I'll add those errors in.
* Thanks I was a bit confused on how the transformation matrix played a part, but the last part about z length; to which are you referring to? Z probe trigger height or the actual Z length (maxPos) of the gantry? 
* Alright good point, I'll leave it as is (M323)
* Right, I was thinking M323 could be fed a specific directory whenever used in which to scan in inside of but er, current mcode commands in FW/Host/Server don't support parameters and strings at the same time though right? I noticed forcing repetier server to mark a command as string command always forces it to send a string. In which case I wonder what'd be the best way to handle custom directories. 
* I think M30 would work well for this. Though at the moment it's rmdir can't delete folders with files in them. 

* Ah, no not a deviation. I'll include a GNU license header. I haven't worked with the AVR flash functions much. I'll quickly read up on them and fix the lib for AVR.

repetier:
> * Do you mean as a non-silent handle EEPROM value? ATM there's a print for the bed temp appended to reportDistortionStatus() so G33 L reports it

Yes, thats what I meant. Also adding it would make it editable which is not so good also user does it on own risk then.

> * I'll add those errors in.
> * Thanks I was a bit confused on how the transformation matrix played a part, but the last part about z length; to which are you referring to? Z probe trigger height or the actual Z length (maxPos) of the gantry?

Motion1::maxPos[Z_AXIS]  is meant here. Won't change much but with thin layers could matter.

> * Right, I was thinking M323 could be fed a specific directory whenever used in which to scan in inside of but er, current mcode commands in FW/Host/Server don't support parameters and strings at the same time though right? I noticed forcing repetier server to mark a command as string command always forces it to send a string. In which case I wonder what'd be the best way to handle custom directories.

That is a good point. Actually you can have parameter at least in server if they are before the M code. Logic is to parse and as soon as we detect text command the rest is text. The ascii parser in firmware also would handle it that way. But host in ascii mode e.g. would change order of parameter, so apart from being confusing it would be more safe to not rely on it and stay with one text parameter.

One solution would be having a default directory and allow M374 to just change directory and M323 could import/export to current selected directory. M375 would then be removed. Also I think you should set temperature in M323 to load and not use bed temperature.

> * I think M30 would work well for this. Though at the moment it's rmdir can't delete folders with files in them.

A test could be added and then recursively delete it, but that is also a bit dangerous in case of an error. So maybe just allow deleting empty directories.

> * Ah, no not a deviation. I'll include a GNU license header. I haven't worked with the AVR flash functions much. I'll quickly read up on them and fix the lib for AVR.

Great. Easiest way is to copy to char array and use it as currently when being called with PSTR. See here:

```
void GUI::bufAddStringP(FSTRINGPARAM(value)) {
    while (bufPos < MAX_COLS) {
        uint8_t c = HAL::readFlashByte(value++);
        if (c == 0)
            break;
        buf[bufPos++] = c;
    }
    buf[bufPos] = 0;
}
```

AbsoluteCatalyst:
Got some time in to quickly write up the changes.

> One solution would be having a default directory and allow M374 to just change directory and M323 could import/export to current selected directory. M375 would then be removed. 

I implemented this solution, however just a slight variation keeping the arbitrary filename importing/exporting of m374/m375.

Since the M375 import cmd was text already, if you now feed it a valid directory path instead of a filename (Ending in "/") it'll change the auto import directory for the M232 S2.

Eg. 
M375 matrixes/
12:09:40.469: Bump matrix auto-import directory set to matrixes/
M323 S2 X80
12:24:25.406: Scanning for matrixes in matrixes/ directory...
12:24:25.417: Importing bump matrix file: 80bed.csv...

Think this works good?

I did want to try keeping the ability to set arbitrary file names for these since occasionally the user may instead have weird configurations that'd need to name/id/remind them differently, it'd also keep in track with the gcode wiki's M374/M375 commands. 

Also as suggested, I dropped the M232 current bed target temp idea and went with a X<temp> parameter which makes more sense.

repetier:
Sounds good to me. Already added M374/M375 also for next server release to be known as text commands.

AbsoluteCatalyst:
That was fast! Er one last second change. I realized I left the starting default ram directory string for the auto import as root dir. Think it'd be better suited as a macro in the config? BUMP_DEFAULT_AUTO_IMPORT_DIR?

repetier:
> That was fast! Er one last second change. I realized I left the starting default ram directory string for the auto import as root dir. Think it'd be better suited as a macro in the config? BUMP_DEFAULT_AUTO_IMPORT_DIR?

Yes, along with a catch in repetier.h for the case it is not defined. I try that as much as I can to give missing macro values a good default if missing. Makes updating from older versions much easier.

AbsoluteCatalyst:
> Yes, along with a catch in repetier.h for the case it is not defined. I try that as much as I can to give missing macro values a good default if missing. Makes updating from older versions much easier.

Ah, I put it alongside the GRID_SIZE missing check in levelingmethod.h since I saw the many of the bump/leveling macros were tested there

repetier:
> Ah, I put it alongside the GRID_SIZE missing check in levelingmethod.h since I saw the many of the bump/leveling macros were tested there

Also good as long a sit gets catched somewhere. So is it ready to do final merge?

AbsoluteCatalyst:
Yeah everything seems good to go. If I find any issues with it i'll push a fix.

PR #930, opened 2020-05-18
Correct Rumba32 TEMP_x_PIN defines to match  PinMap_ADC

schrod:
The HAL::analogEnable function maps the TEMP_x_PIN  to an analog channel using PinMap_ADC which uses PinNames enum from STM (PC_0) rather than the Arduino defines (PC0).


repetier:
Strange that I got so far correct temperatures, but I think you were right.

PR #932, opened 2020-05-23
feature IO_BEEPER

AbsoluteCatalyst:
This updates beepers to match the IO module system and also implements the themes concept for config.io. Most of the file changes are just updating macros and removing the previous beeper code.

I've added a HAL function (HAL::setHardwareFrequency) to allow pwm frequency changes at runtime. It's implemented for the Due, however I don't currently have a SAMD or STM32F4 board on hand to safely implement the function for those processors sadly.

An example of a config.io is:
```
IO_OUTPUT(IOBeeperMain, BEEPER_PIN)
BEEPER_SOURCE_IO(MainBeeper, IOBeeperMain)
```
or: 
```
IO_PWM_HARDWARE(PWMBeeperMain, BEEPER_PIN, 100)
BEEPER_SOURCE_PWM(MainBeeper, PWMBeeperMain)
```
creating themes: 
```
(Name, TONES({Frequency, Duration}, ...)) 
TONE_THEME(ButtonClick, TONES({ 8000, 10 }))
TONE_THEME(ButtonReset, TONES({ 6500, 50 }, { 7500, 80 }))
TONE_THEME(Notify, TONES({ 8500, 100 }, { 9750, 150 }, { 8500, 100 }, { 9750, 150 }, { 0, 500 }))
TONE_THEME(NotifyFail, TONES({ 8500, 80 }, { 0, 50 }, { 8500, 150 }, { 0, 1500 }))
TONE_THEME(NotifyConfirm, TONES({ 9250, 70 }, { 9500, 100 }))
```
The tone arrays are saved in PROGMEM and then later extracted using memcpy_P. Tested it under arduino and it functions without any trouble, so transitioning should be smooth later. 
Frequencies of 0 still act as delays for the queue.

To use the themes there's two macros: 
```
(Beeper Source, Theme, Blocking)
PLAY_THEME(MainBeeper, NotifyFail, true)
```
A direct one shot macro to play the theme that can be applied anywhere within the firmware. (Good for the event system)
Blocking disallows any other tones to play/queue while it's playing.

```
(Beeper Source, Condition, Theme, Play Times)
TONE_THEME_COND(MainBeeper, GCode::hasFatalError(), Notify, 5)
```
A conditional similar to lights io. 
If you set the play times to 0, it'll play the theme endlessly. 
Conditions are blocking by default. 

You can have as many beepers as you'd like. It wasn't really my original plan but allows an interesting use-case of having another louder, stronger alarm/beeper for fatal errors or decouples.

There's HAL code to handle simultaneous IO beeps if they occur however it's not inherently perfect as it tries to find a suitable frequency for the timer in which both frequencies will work, which sometimes results in crazy values. Crazy frequencies simply get clamped to 100khz at worst. I don't believe it'll ever really happen that often regardless.


repetier:
Just want to let you know that I haven't forgotten this request. It is just big so need some time to understand and extend missing parts. I think when server is released I can come back here and merge it.

AbsoluteCatalyst:
No worries and thanks for that update. Looking forward to a new server release.
In the meantime, this gave me a good bit of time to revise the code and catch a few issues. I've updated the pull request with newer commits. There's a few as I split up the commits to make it easier to keep track of the changes.

The biggest changes were: 
- I created a proper implementation file for class methods as the header was getting quite cumbersome. Especially if later a SPI/I2C beeper (some displays?) is added.
- I moved the theme conditions from the periodical timer to the 100ms timer and improved their order/execute handling a little.

As for issues, one that was of note: I found a pretty nasty bug in my due's HAL sethardwarefrequency function caused by a bug within the libsam library itself. PWMC_ConfigureChannel always redundantly sets/clears channel 0's CMR to 1. That ends up also activating whatever's on that PWM channel. 
Only caught that when setting up a different due printer's hardware config and one of my PWM heaters ran away. Glad I got the chance to spot it now.

repetier:
Server is out but there are some bugs in changed features so need to fix them. Guess next week I can check on the beeper stuff.

Great catch of the PWMC_ConfigureChannel problem. Would be really hard to find remotely if a client would get such a problem.

AbsoluteCatalyst:
Oooh. Thanks for the heads up on the release. Great update, wanted a lot of these additions/changes.

For the beeper hal extends, I might be able to implement it for the STM32F4 HAL later on too.

I purchased a SKR E3 Mini v1.2 (STM32F1) recently and copied over the F4's HAL with the intention of adding support for it towards V2.

The changes needed to make it successfully compile were really minimal. (flash handling, small DMA changes, ADC channels) 
Though completely untested atm as I'm actually receiving the board sometime later today. 

(as a side note It's use of software uart for it's built in TMC2209's may pose a problem in general...)

Might not even be worth having a seperate HAL and just #ifdef the changes depending on chip I think.
Will test it thoroughly before anything though.

repetier:
Keep me posted on the STM32F1 - I also think that the STM32F are all quite similar. Main difference is which timers/adc channels they have I think.

I just got a Bigtree SKR1.3 with TMC2209 where they work out of the box as they say. Once I have LP1768 support I will try that as well. But uart is in deed a problem. You can but 2 in a row I think but not sure exactly how that is going to work. Will first try with marlin as that is contained firmware and then teach it the right firmware:-)

AbsoluteCatalyst:
Will do. 
And you're right, quickly looking at the datasheet I see that two skr 1.2 "uart" pins (Z and E axes) sit on two HW remappable uarts (UART4, USART3). Remapping those pins individually then setting them into half duplex mode *should* work I think... 
Might be some catch in there we'll have to find out though. 
Should be fun.

repetier:
I just registered that while you have provided a "ButtonClick" theme we have no sounds in the menu but a option to disable sounds:-) In V1 we always had clicks on next/select. I think I should reintroduce these sounds. Currently I think we need 3 sounds:
- nextPrevClick
- buttonClick
- errorClick (reset / error screen)

Now it would be a pain if everyone needs to add these manually on the other hand I'd also like to have them overrideable. The only idea I currently have is to make a macro variable
#OVERRIDE_DEFAULT_TUNES 0
which when set to 1 would not define the default tunes and would rely on the tunes defined with given names in configuration_io.h instead.

What do you think about using default tunes for standard actions and do you see a better solution for overriding?

AbsoluteCatalyst:
Bringing back default/stock tones for the UI would be good.
Can't seem to think of any sort of automatic override system that'd work at compile time though... Maybe with some SFINAE tricks but it might get complicated. Think the override flag is fine for now

I was thinking that vendors could define their personalized printer's/board tones as well should they want to.

repetier:
Ok, I needed to make some more changes to your solution.

The control in periodicalAction was a bad choice as I learned. With my slow graphic display it gets not called frequently enough so next/prev became a quite long sound. So we needed a more frequent solution and I have now chosen to always activate servo timer if beepers or servos are defined. That function gets called 400 times per second so a good compromise between precision and extra load.

Sounds in menu now are also implemented and there is a new function 
Printer::playDefaultSound(DefaultSounds sound)

That plays the sound and is extendable if required to add more themes.

STM32F4 worked good, just could not make the hardware timer - the pin for my beeper has no hardware pwm available.

repetier:
While working on the documentation I think I found a bug in the cond handling. It will play the wrong theme/count if you have multiple conditions.

If you check for
C1
C2
C3

and C3 becomes true first is plays it.
If now C2 comes true as well curConditionStep increases and you play the theme of the last true condition, which is C3 and not C2. When C1 then becomes true you play C3 again.

There is no real solution to this I think unless you have a state variable if a condition was already played. So it must be more like

```
#define TONE_THEME_COND(name, source, cond, theme, playTimes) \
    static bool name = false; \
    if (cond && name != cond && !source.isMuted()) { \
        if(source.setCondition(theme, playTimes);) { \
        name = cond; } \
    } else name = false;

```

I  know you would need to change setCondition but not sure in your system how to handle the playTimes problem. Playing it once would be just adding it in queue if there is space. So that is another problem to consider here. Would be glad if you could ook into this problem.

You can use debug settings to play sound like Printer::debugDryrun() - you also have debugEcho and debugErrors so you can test with M111 easily.

AbsoluteCatalyst:
Thanks for spotting that. I did have trouble trying to get the conditions to properly keep states logic wise and thought it was in the clear. I think I'm going to handle them as small objects/structs that keep track of themselves this time around instead. 
I'll definitely look into it tomorrow.

AbsoluteCatalyst:
Okay! I've got a working update.
In the end, no, I couldn't seem to find any real solution to come up with to avoid conflicts like this besides having another parameter as a unique key/name for the condition macros and handling them individually/as objects/states.
I originally tried to avoid the extra parameter because I felt it might confuse the end-user. But I don't think it'll be too much of an issue? 

TONE_THEME_COND(name, source, cond, theme, playTimes) like your example

The conditions have their own short structs containing how many times they've been played/if heard/play count/loop. 

It shouldn't change/restart the last theme's play count if another one plays anymore. The debug flags highlighted that curConditionStep wouldn't update properly sometimes.

I fixed the problem you mentioned with handling playCounts. Now It'll fill the buffer then continue filling it every runConditions until the playCount is properly met.

I've made one intentional change now that each condition has their own storage (Could be removed): 
If we're inside a higher hierarchy condition -> and a lower one triggers that we've never heard before -> it'll play that lower condition as we exit the higher condition **if** it's still true. 
(eg. keeps track if they've been "heard" or not.)

I've been testing this with no issues so far:
```
TONE_THEME(beep1, TONES({ 3000, 100 }, { 0, 500 }))
TONE_THEME_COND(CondIndex1, MainBeeper, Printer::debugEcho(), beep1, 2)    // 1
TONE_THEME_COND(CondIndex2, MainBeeper, Printer::debugDryrun(), beep1, 4)  // 8
TONE_THEME_COND(CondIndex3, MainBeeper, Printer::debugNoMoves(), beep1, 8) // 32
TONE_THEME_COND(CondIndex4, MainBeeper, Printer::caseLightMode == 0, beep1, 16)
```
Hopefully this is the end of anymore trouble.
I'll post a pull soon as I've been needing to merge from a few commits back.

repetier:
I think there is no way to avoid the extra storage. You can not cover n conditions/orders with less then n variables without possible faults. n is the minimum. 

Also it is in line with all other objects all having names as well, so no problem there anyway I'd say.

PR #935, opened 2020-06-27
IO_Beeper condition conflict fix

AbsoluteCatalyst:
https://github.com/repetier/Repetier-Firmware/pull/932#issuecomment-650559664
Conditions with the same theme shouldn't conflict anymore, nor should the last theme be repeated more than it should.
Conditions now correctly play all the way until playCount is reached. (By refilling the buffer when needed)
Conditions that were never "heard" but are still valid now play after leaving a higher ranked condition.
Sanity check for themes larger than the beep buffer

(Now needs a unique name)
TONE_THEME_COND(name, source, cond, theme, playTimes)


repetier:
Thanks. Looks good.

PR #936, opened 2020-06-27
Add PT1000 temperature tables

AbsoluteCatalyst:
For 4.7k and 1k pullup resistors.


repetier:
What is the resistor here? Aren't they used on special chips returning x mV/°C and no resistor involved? Just asking for the documentation.

AbsoluteCatalyst:
The resistors are in series. PT1000's have the same setup as generic thermistors but are much more accurate/survive higher temps. No IC needed for these. They're also less expensive than PT100's.

PR #937, opened 2020-06-27
Heater name with decouple errors

AbsoluteCatalyst:
printName addition to decoupled fast rising/holding reports
Small language fixes/changes


repetier:
Here you see I'm no native english speaker :-)

PR #939, opened 2020-07-25
Small general V2 fixes

AbsoluteCatalyst:
Hey, I have a few pulls I've been meaning to do for dev2

This one is just really small syntax/cpp fixes: 
- added constexpr to IO_INPUT's pin()s (for constexpr functions etc)
- set tool.h's base updateDerived pure. Linker issues under some circumstances
- HAL's were missing INPUT_PULLUP option
- redid some initializer syntax for light source pwm and fans
- added a sanity check for measure() for bad eeprom configurations  

Let me know if any of these cause any issues


repetier:
Looks all good.

PR #943, opened 2020-07-25
Z-Probe check if probe never triggered and fix secureDistance

AbsoluteCatalyst:
I updated secureDistance to just be getBedDistance() * 1.5 in order to generate shorter maximum travel distances. There's also now a check for stepsRemaining to see whether or not the probe was ever triggered. This should mitigate some damage in case we have a bed crash, since it'll error out fairly quick.

I don't believe this secure distance value should have an issue with z max homing printers from what I can see? Though If anything, I can quickly change it. 

_Edit_: I noted in my original commit that I thought I found a small bug in stepsRemaining wherein rarely; a previous value would remain and then affect/append the next stepsRemaining value giving us erroneous results. 
After examining/debugging it again today, I can't seem to reproduce what I originally saw. My mistake.

### DIFF
--- a/src/Repetier/src/drivers/zprobe.cpp
+++ b/src/Repetier/src/drivers/zprobe.cpp
@@ -21,6 +21,7 @@
 
 #include "Repetier.h"
 
+#define Z_CRASH_THRESHOLD_STEPS 50
 #if Z_PROBE_TYPE == Z_PROBE_TYPE_DEFAULT
 
 uint16_t ZProbeHandler::eprStart;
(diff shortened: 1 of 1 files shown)


repetier:
I see what you want to try, but this is bad for same use cases. All high level functions using beddistance as start height would work fine, but sometimes you use a G30 at your current z height and then the distance would be not high enough. Also G32 uses ZProbeHandler::optimumProbingHeight().

One idea is using Z but that only works after z homing and when bed moved down (I have a printer where it can happen) it will also never touch the bed. Also in that case probing makes no sense except again with G30.

I think we can use it with an added test and reduce it only when z is close to ZProbeHandler::optimumProbingHeight().

AbsoluteCatalyst:
Thanks for that, I hadn't realized I broke the G30 command with this. Whoops!

repetier:
No problem. That is one reason I always read code changes before merging them. Sometimes I see things not obvious or think about side effects form my long experience with firmware. The good thing with your requests is that you always try to cover also the devices not used by you and your code has good quality. So always a pleasure to merge them.

AbsoluteCatalyst:
Thanks for the kind words. Happy to be helping out!

PR #945, opened 2020-07-25
Z-Probe display controller configuration menus 

AbsoluteCatalyst:
Added probe speed, bed height, trigger height, heater pausing.
The default/bltouch probes have coating height and x/y offsets. 
The nozzle probe has just nozzle temperature.

(I thought of just doing a #if PROBE_TYPE != NOZZLE_TYPE for the coatings/offsets, but that might defeat the purpose of their own config functions I think)
Probe delay might be another option to add, especially for the bltouch.

I removed EEPROM::markChanged from all the setZProbeHeight's and placed it in the G30 call instead. These menu's don't call markChanged, just like the other config menus.

Now just needs a control menu.


repetier:
Nice additions. I saw that we have coating for BLtouch. I'm thinking of removing it.  Reason for coating are inductive sensors measuring through coating so there it would make sense, but bltouch is physical so that is unneeded and only adds confusion. What do you think?

AbsoluteCatalyst:
Agreed, it makes sense to remove it for the bltouch

repetier:
While removing I also saw it was never meant to be there. No eeprom entry for it anyway:-)

PR #948, opened 2020-07-30
Make a bare G92 reset coordinates.

rasilon:
Per https://reprap.org/wiki/G-code#G92:_Set_Position
    "A G92 without coordinates will reset all axes to zero on some
    firmware."

My specific use case was so that I could easily repeat parts in Laser
mode, using generated G code that assumed absolute coordinates, knowing
the origin of each part.


repetier:
Must have missed the mail for this.

PR #956, opened 2020-08-30
define SD_SPI_SPEED_MHZ for config changes to SD SPI speeds

AbsoluteCatalyst:
PS. What's the dummy init for?


repetier:
I think it was because some boards had problems with first attempt. Maybe due to instabilities during power up or because sd card took longer to get stable after sd detect. So maybe the delay it self would also have helped. But too long ago to remember exactly or retest.

AbsoluteCatalyst:
Ah okay. Perhaps to help SW-SPI at boot as well, etc.

PR #958, opened 2020-08-31
ZProbe - Fix bltouch alarm loop

AbsoluteCatalyst:
Small find If the user's bltouch isn't wired up properly, or accidentally unplugged.


repetier:
Good idea also this should only be called when we are at old start position. But who knows when the bltouch will fail.

PR #960, opened 2020-08-31
ZProbe - Control menu 

AbsoluteCatalyst:
I've added the control menu for ZProbes though at the moment it's quite small with only two options (Run Single Probe, Run Autoleveling) unless you're using the BLTouch. +(Deploy, Stow, Alarm Release, Self Test).

I first used Leveling::measure to run the Autoleveling command and runProbe for the Single Probe,
But I saw that there's routines for eg. ZMax printers that need to run after measuring at the GCode_ functions. 
In effort to run those correctly as well, I'm calling Commands::executeGCode with a GCode instance:
```
GCode innerCode;
innerCode.source = GCodeSource::activeSource;
innerCode.internalCommand = true;
innerCode.setG(32); // G32
Commands::executeGCode(&innerCode);
```
Are you alright with me accessing/manipulating the GCode class/functions like this? Or should I look for another route?

### DIFF
--- a/src/Repetier/src/controller/menu.cpp
+++ b/src/Repetier/src/controller/menu.cpp
@@ -222,49 +222,47 @@ void __attribute__((weak)) menuConfigProbe(GUIAction action, void* data) {
     GUI::menuEnd(action);
 }
 
-void menuRunProbeOnce(GUIAction action, void* data) {
+class GCode;
+void __attribute__((weak)) menuRunProbeOnce(GUIAction action, void* data) {
+#if Z_PROBE_TYPE != Z_PROBE_TYPE_NONE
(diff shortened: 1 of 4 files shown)


repetier:
Solution is ok also a simple call directly to the gcode function would be easier:

```
GCode innerCode;
innerCode.source = GCodeSource::activeSource;
innerCode.internalCommand = true;
GCode_32(innerCode);
```

Also I guess seeing this I will make a extra constructor for internal gcodes so this could be reduced and to make it easier to use it in similar cases. But you can leave this for now. I will then change it along when I add the constructor.

AbsoluteCatalyst:
Sweet, alright. I had used the GCode function at one point, though I was worried about accessing it without going through command's activeSource handling portion in case I missed something somewhere

PR #961, opened 2020-09-06
STM32F4 Compile fixes/Compatibility with STMDuino 1.9.0/STM Platform 8.0.0

AbsoluteCatalyst:
Hey, I saw some people mentioning the rumba doesn't compile at the moment due to the newest platform update.
I've been working on the F1's HAL under that version, so here's some changes to help make F4's HAL compatible as well.
If you've already patched it on your end then you can delete/deny this pull, no worries. 

The main change needed was the removal of the HardwareTimer* parameter for timer callbacks, and the new usage of refresh(); since timer overflow preloading is enabled by default now. 
(refresh immediately updates the overflow/compare register. Needed if overflow was set to 0 by any chance.)
The platform itself needed to be updated to 8.0.0 as well otherwise we'd get a cmsis compile error. 

I don't own a rumba32, but I do have my hands on a Nucleo-64 (STM32F446) in which it seems to run fine under
Though double check whenever you get a chance to make sure everything is hunky-dory of course :) 

Other changes in this pull are:
The USBD_PID macro addition you mentioned on your forums. 
An extra check for frequency before initializing pwm outputs or changing frequency.
Removal of pause/resume for pwm changes/frequency changes and SW beeper (beeper manipulates overflow preloading for slightly improved smoothness) 
Fixed small typo in importEEPROM (sdcard variant)
I'm using exact package/framework versioning in platformio.ini, this is probably unneeded.


repetier:
Thanks for being ahead of me:-) Need to fix new server first since it contains some ugly problems.

Like the solution with forcing the stmst version. Solves the problems with users still having the old one as I have.

PR #962, opened 2020-09-07
PWM Light sources remove unneeded virtual declarations 

AbsoluteCatalyst:
Realized some of these high speed functions in IO_Beeper/IO_Lights i wrote a while back were never getting inlined/generating expensive vtable lookups. I don't think these three functions here needed to be virtual for now (IIRC)


repetier:
Yes, no need to be virtual. Only methods of the base class reused should be virtual and marked with overwrite.

PR #964, opened 2020-09-13
Async display init + skippable intro + show reboot warnings

AbsoluteCatalyst:
I've split the init for both of these displays so that a new function (processInit()) is called during boot-up every GUI::update(). Should help displays that might require a slightly longer delay without slowing down boot.

I later extended this slightly to let us skip the intro screen if we attempt to scroll/click while it's animating.
Three gui boot states: 
```
GUIBootState::DISPLAY_INIT
GUIBootState::IN_INTRO
GUIBootState::READY
```
(Active/current state stored in GUI::curBootState)

The final thing was an addition to our HAL's to store our last boot reason; and displays a warning if we've had a watchdog reset or brownout. The reboot enums are: 
```
enum class BootReason {
    SOFTWARE_RESET = 0,
    BROWNOUT = 1,
    LOW_POWER = 2,
    WATCHDOG_RESET = 3,
    EXTERNAL_PIN = 4,
    POWER_UP = 5,
    UNKNOWN = -1
};
```
defined in repetier.h.
Current  boot reason is stored in HAL::startReason.

One small thing: I hadn't noticed that SAMD's could return multiple reasons when I began with the other HAL's. In this case i've ordered the checks for SAMD by priority-ish for now.

### DIFF
--- a/src/Repetier/src/Repetier.h
+++ b/src/Repetier/src/Repetier.h
@@ -206,6 +206,16 @@ public:
     virtual void executeGCode(GCode* com);
 };
 
+enum class BootReason {
+    SOFTWARE_RESET = 0,
+    BROWNOUT = 1,
+    LOW_POWER = 2,
+    WATCHDOG_RESET = 3,
+    EXTERNAL_PIN = 4,
+    POWER_UP = 5,
+    UNKNOWN = -1
+};
+
 #include "io/temperature_tables.h"
 #include "Configuration.h"
 
--- a/src/Repetier/src/boards/SAMD51/HAL.cpp
+++ b/src/Repetier/src/boards/SAMD51/HAL.cpp
@@ -43,6 +43,7 @@ extern "C" char* sbrk(int i);
 char HAL::virtualEeprom[EEPROM_BYTES] = { 0, 0, 0, 0, 0, 0, 0 };
 bool HAL::wdPinged = true;
 uint8_t HAL::i2cError = 0;
+BootReason HAL::startReason = BootReason::UNKNOWN;
 
 volatile uint8_t HAL::insideTimer1 = 0;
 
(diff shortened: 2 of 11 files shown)


repetier:
Good idea. But I think the status message of brown out/watchdog might not be enough when you are connected with e.g. Repetier-Server. User would eventually look at display and see the IP number form server instead of the message. I think pushing the warning screen when intro screen is finished would be better as it is a wizard waiting for "ok" to continue. So instead of
            if (HAL::startReason == BootReason::WATCHDOG_RESET) {
                GUI::setStatusP(PSTR("Reset by Watchdog!"), GUIStatusLevel::WARNING);
            } else if (HAL::startReason == BootReason::BROWNOUT) {
                GUI::setStatusP(PSTR("Brownout reset!"), GUIStatusLevel::WARNING);
            }
we would use something like
push(warningScreenP, PSTR("Reset by Watchdog!"), GUIPageType::STATUS);

where we need to write warningScreenP - warningScreen requires a ram address and status might already be overwritten by ip. I can write that.

What do you think about that?

AbsoluteCatalyst:
Oh that's how it's set up now, it's a warning popup "ok" prompt right after the intro

AbsoluteCatalyst:
Oh don't we also print out the reboot reason on printerSetup too? A small warning prompt on repetier server itself would be quite cool too!

repetier:
Yes we write it and I agree having it as information on server might be in deed a good idea. Especially the watchdog/brownout but that is hard since all firmwares have it different and might change names. But as info on the "Connection information" screen it might be nice if known.

I then make the change for the warning screen and put it there.

repetier:
Ok have made the changes now we get a warning screen in that case.
Also found a bug on start screen delay that happened on some encoder positions where a next triggered contentChanged early.

AbsoluteCatalyst:
Sweet thanks for that fix. Bug was elusive on my end/screen and couldn't tell if it actually existed or if I moved my encoder myself by accident.

PR #965, opened 2020-09-18
Add destination checking print

AbsoluteCatalyst:
Small message notification if called without any XYZ coordinates.


repetier:
I'm just wondering why it is P disabling the check. On V1 it was S and in main.cpp docs it still is S.

On the other side I saw that RepRapFirmware uses S for laser intensity which would also be a good idea to make gray levels possible with less gcodes. And then we would have the same code for it.

AbsoluteCatalyst:
I'm not entirely sure, I did also remember it being S too. Maybe it was used for debugging at some point and forgot? (Did I accidentally change it?)

repetier:
I rechecked it again. Now found I already used S for laser intensity already so using P was a decision I made already to make S for laser intensity or any other secondary tool value.

PR #967, opened 2020-09-18
Feature Hide baudrate config if only connected to nativeUSB

AbsoluteCatalyst:
I added/implemented grabbing the USB host into the new GCodeSource class, and now we track a pointer to the usbHostSource if we have one. 
If we only have a native USB connection this hides the baud configuration option from the eeprom and menu since baud rate doesn't apply to native. (well, technically it still gets passed down from the host actually, but doesn't do anything with it)
It's purposely _only_ hidden in case the user adds another serial source (TFT, bluetooth etc) it shouldn't corrupt eeprom nor lose their settings.

Unfortunately the way i'm detecting the usb host is a naive-ish pointer compare with the "SerialUSB" class since there's no API available. The class seems pretty consistent across arduino platforms though, I don't think it should cause trouble. 
I did #if it out for AVR platforms, since I don't think any AVR printer boards with native USB actively exist. (?)

PS. Do we have any MCode Command sent/official way to detect Repetier-Server/Repetier-Host instances from the firmware itself? :) (M539?)
It'd be useful to have a flag for it as well when connected I think. More server-firmware integration features like host rescue are cool.

### DIFF
--- a/src/Repetier/src/PrinterTypes/Printer.cpp
+++ b/src/Repetier/src/PrinterTypes/Printer.cpp
@@ -524,13 +524,7 @@ void Printer::setup() {
 #endif
 #endif
 
-#if (defined(BLUETOOTH_SERIAL) && BLUETOOTH_SERIAL > 0) || CPU_ARCH == ARCH_AVR
-    constexpr bool isSerialUSB = false;
-#else
-    // GCC's __builtin_strncmp works with constexpr
-    constexpr bool isSerialUSB = (strncmp(PSTR(VALUE(RFSERIAL)), PSTR("SerialUSB"), sizeof("SerialUSB")) == 0);
-#endif
-    Printer::setNativeUSB(isSerialUSB);
+    Printer::setNativeUSB(!GCodeSource::hasBaudSources());
 
 #if SDSUPPORT
     //power to SD reader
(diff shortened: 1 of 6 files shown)


repetier:
Agreed that baud rate for native is not necessary.

We have M530 S1 to know we are printing from host/server. But there is in deed no gcode telling firmware it is handled by a host/server. On the other side if we receive communication we know it comes from a server.
M539 S1 is send by server so firmware knows it is connected to server that supports RequestContinue:/RequestPause:/RequestStop: also this flag is never tested as I see.

What do you have in mind?

AbsoluteCatalyst:
Nothing inherently big or fancy at the moment; I was thinking about the IO system in the firmware and peripherals shared between the printer and server. 

I saw you made a general use SPI IO module, which opens up a lot of possibilities. I have a concept I was toying with which was "IO Sensors" too, and was planning to hook on a BME680 with it. Sharing the sensor data directly with the server to use and display using an optimized protocol would be quite good. 
I'm not sure how much more we can fit with the current autoreport format we use? 

There's also that piezo probe coming out now for example that accepts I2C/UART. 
An idea I had was being able to do a firmware update on it remotely through the server, then literally passing it through the printer firmware, to update the probe. 

Another thing was perhaps a optimized protocol for logging firmware misconfigurations, errors, diagnostics etc on the server itself, since we have memory/flash limitations. For example the firmware detects a unstable thermistor connection during a print, it gets logged and RS warns the user at the end that they might have a wiring harness going faulty. etc.

PR #969, opened 2020-09-18
Fix HW TMC Stepper baud & IRQ priorities

AbsoluteCatalyst:
I was trying to figure out why my TMC2209's had connection issues/dropouts on my main due printer with HW-USART. 
Never noticed serial was creating interrupts at 60KHz-ish, and at priority 0.
( I could have sworn the serial classes defaulted to 15? ) 
I adjusted the TMC2209's to 115200 baud and set their priority to 5. (I also set the TMC08's priority to the same)
No more connection issues, and my steppers seem slightly quieter/smoother. 
Think this default baud/priority should be good for everyone?

### DIFF
--- a/src/Repetier/src/io/io_stepper.h
+++ b/src/Repetier/src/io/io_stepper.h
@@ -127,9 +127,11 @@
     name.eepromReserve();
 #define STEPPER_TMC2208_HW_UART(name, stepPin, dirPin, enablePin, serial, rsense, microsteps, currentMillis, stealth, hybridSpeed, fclk, minEndstop, maxEndstop) \
     serial.begin(115200); \
+    serial.setInterruptPriority(5); \
     name.eepromReserve();
 #define STEPPER_TMC2209_HW_UART(name, stepPin, dirPin, enablePin, serial, rsense, microsteps, currentMillis, stealth, hybridSpeed, slaveAddr, stallSensitivity, fclk, minEndstop, maxEndstop) \
-    serial.begin(500000); \
+    serial.begin(115200); \
+    serial.setInterruptPriority(5); \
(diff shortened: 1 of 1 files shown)


repetier:
I knew about the priority. This is because buffer is only 2 byte I think so new chars need to be polled fast to not get communication errors. But I thought the interrupt would be triggered if a byte is complete and not at 60KHz. Why do you think that? Sounds like a waste of cpu time.

AbsoluteCatalyst:
That's what I thought as well. It was coincidence that I was hunting down IRQ's/performance drains and decided to time that IRQ in my due using DWT/CYCCNT. The UART IRQ was triggering every 15-16us~ or so. Let me quickly double check since I have it infront of me sec.

AbsoluteCatalyst:
Okaayy bossa finally let me upload my firmware, anyway at the moment we're at 115000 baud. 
I'm getting around 8533 clock cycles between interrupts it seems. I run my due a little faster at 102mhz, so it's 9.8ns per cycle.
and that puts it at 12Khz. still a little bit fast tbh.

repetier:
Not sure if it so much off. 115200 / 12000 = 9,6 bit per interrupt. We communicate with 1 stop bit so 9 bit is what is to expect. So yes there is some error of 0,6 bit/interrupt. But that makes much more sense already then 60KHz.

AbsoluteCatalyst:
Ahhh, you're right. I do wish they'd had implemented DMA with it back then. 
Yeah at those speeds the TMC's were struggling to keep up too. I shouldn't have cranked it up to their max in the beginning. 
I had a really rare issue sometimes on boot-up where their microstepping settings were wrong and set to 256. I think that was my cause.

PR #970, opened 2020-09-19
BLTouch configurable deploy/probe delay

AbsoluteCatalyst:
Added this as a setting for the eeprom and menus. Added a hint for the macro in config.h as well.


repetier:
Good idea. Never  thought about is always assumed it would be fast enough until first touch. But if you start close it is a edge case where it is important and I guess it is also enough to trigger alarm in case you start too low.

PR #972, opened 2020-09-22
GUI U8G2 - Offset all text 1px away from borders

AbsoluteCatalyst:
The status bar text was riding on the edges of this U8G2 display and it looked a little odd/uncentered.
This adds a 1px of margin to it. Looks perfect now. 

Moved any text at X = 0px to X = 1px
Moved the top status text down just 1 px
(Line kept at the same height.)

Edit: Actually, is this too subjective or perhaps an issue with just my screen's very visible pixels? 
(I do really love the always-on status bar concept/setup, by the way.)

IMG_20200921_214900_01

### DIFF
--- a/src/Repetier/src/controller/drivers/DisplayU8G2.cpp
+++ b/src/Repetier/src/controller/drivers/DisplayU8G2.cpp
@@ -148,7 +148,7 @@ void __attribute__((weak)) drawStatusLine() {
     GUI::bufAddHeaterTemp(heatedBeds[0], true);
 #endif
     lcd.setFont(u8g2_font_5x7_mf);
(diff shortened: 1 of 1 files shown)


repetier:
I'm a bit unsure here. On my screen the difference is nearly impossible to see so that it makes no real difference.
What makes a difference is that we then hit the right side 1 pixel earlier. With 5x7 font we use 125 and this no full pixel. With 6x10 we use 126 pixel with 21 chars so adding one pixel makes it centered. So for the better look of your display I take a push request:-)

AbsoluteCatalyst:
Ah I see, thanks!

PR #973, opened 2020-09-22
IO_Beeper optimizations & volume handling

AbsoluteCatalyst:
This was meant to go along with the io light fixes i sent in a little while back. 

It never came across me that we could just vary the duty cycle to change the volume on these. 
It's not 100% linear though depending on the physical properties of the hardware, but so far actually works. 

Software beepers as well. (Even this active beeper on this U8G2 display.)

EEPROM still uses a byte since it's a 0-100% value, so it shouldn't cause any trouble for anyone switching over.

The SAMD51 unfortunately doesn't have it yet as I'm unsure on how to set up the capture compare on that platform. It just still functions as normal. Should be extremely easy to set up when using the due as a reference however. 

It's set for the due hal, stm32f4 hal, and stm32f1 hal i'm about to pull.


repetier:
Interesting idea. Like it to change volume. Curious on how much difference I get on my display beeper.

PR #974, opened 2020-09-22
STM32F1 SW SPI U8G2 Displays

AbsoluteCatalyst:
_I split this separately from the HAL since it's a bit iffy._ 

Do you remember this? https://github.com/olikraus/u8g2/issues/749
For the STM32F1 I had to do the same workaround here.
 
I found the display needs a slower send speed while it's booting up to avoid a glitched first few rows, and it appears this display can't handle more than 700khz~ from what i last saw on my logic analyzer.
(I don't remember if the datasheet actually already mentioned that)

I did something experimental here which i'm unsure how it would be handled if i made a pull to u8g2 too. 
(eg is this too specific to *us* only?)

There needed to be a delay between bytes otherwise it'd glitch out since we'd be going too fast, but using nops would occasionally drag out the delay, especially while we're doing more complex movements/functions. 

I'm manipulating CYCCNT/EXCCNT to measure clock cycles spent during + between bytes, and also cycles spent inside interrupts as a form of "processor overhead" (then averaged)

Admittedly I think the numbers used right now to control the delay (clocksPerByte/itOvrheadMult) are a tiny bit "magic".
But I'm managing around 50~ ms per refresh regardless of what else is going on at the time.

As a sidenote I #ifdef the due's code since I think the timing might have been very delicate/specific and I didn't want to break it.
(Do you no longer use it? The only other code was for AVR)

What do you think about this?

### DIFF
--- a/src/Repetier/src/u8g2/U8x8lib.cpp
+++ b/src/Repetier/src/u8g2/U8x8lib.cpp
@@ -551,43 +551,70 @@ inline void u8g2_delay_x100ns(int x) {
         asm volatile("nop" ::); // 11.9ns
         asm volatile("nop" ::); // 11.9ns
         asm volatile("nop" ::); // 11.9ns
-// STM32F1's run at 72mhz, so each is 13.9ns
+                                // STM32F1's run at 72mhz, so each is 13.9ns
 #if !defined(STM32F1)
         asm volatile("nop" ::); // 11.9ns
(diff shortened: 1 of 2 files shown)


repetier:
Yes I remember it. One reason I do not dare to update the library:-) I fear to loose my modifications breaking the displays.

PR #975, opened 2020-09-23
STM32F1 HAL & E3 Mini V1.2 variant definitions

AbsoluteCatalyst:
There's still some more small things here and there needed, but for the most part this is done. (I think) 

The differences between this and STMF4's is mainly manually handling the NVIC timer IRQ functions, among other small things. STM32Duino's callbacks/HAL added a lot of latency, I worked around this by renaming the irq functions in the vector table and stealing them back from the HAL.

I still liked most of the HAL and it's provided functionality, and I didn't want to strip it completely; but I moved a little closer to STM's LL driver at least for this processor.

Oh right and I also did the same for the software serial library in order to improve baudrates and drastically improve reliability. I planned running dma functions for it at a later date when I get some free time. It's pretty solid now. 

I still haven't made a proper ender 3 sample config so i'll be uploading one in the next few days.
(What's a good up to date sample config to work with as a base btw? the main one yeah?)

Let me know if you have any questions!


repetier:
Great work.
What I noted is that upload address is 0x8007000 instead of 0x8000000 which is the default address for STM chips due to a bootloader. I also noted you are using stlink as uploader which is fine for you, but most users will not have one. Does selecting dfu work with the board or does it not have a boot button? And if so would platformio use the correct address?
For the server I can make an extra entry for DFU with that address it it works. Would be easy.

Default config is a good starting point I guess. It is the only one where we consistently add new parameter.

repetier:
Just remebered where the code in platformio is. It is in main.py in stm32/builder

```
elif upload_protocol == "dfu":
    hwids = board.get("build.hwids", [["0x0483", "0xDF11"]])
    vid = hwids[0][0]
    pid = hwids[0][1]

    # default tool for all boards with embedded DFU bootloader over USB
    _upload_tool = "dfu-util"
    _upload_flags = [
        "-d", "vid:pid,%s:%s" % (vid, pid),
        "-a", "0", "-s",
        "%s:leave" % board.get("upload.offset_address", "0x08000000"), "-D"
    ]
```

But unsure if your
`board_upload.offset_address = 0x8007000`
becomes upload.offset_address. But function is called board.get so that might strip the board_ prefix.

AbsoluteCatalyst:
Sorry just got these emails now. 

No this board doesn't support DFU unfortunately, and yeah the offset is for bigtreetech's bootloader. 
It's a little finicky, but it's a sdcard based bootloader. Searches for a firmware.bin on the card, then applies it. 

**Note:** I forgot to put in a warning that it doesn't ever delete the firmware.bin after flashing, so rebooting with a firmware.bin in your sdcard will always reflash, which can be a little annoying

Most users; if not all, will probably be using this method for it. I just used the stlink since it was a faster workflow. Perhaps some users/developers might use the stlink, so I left my upload settings/things in the ini just in case.

Aye, board_upload seems to translate into upload.offset_address. 
There was a platformio wiki page on it but I don't have it on hand. (Some other board_* functions work this way too)
Took a little while to figure out how to get both the firmware.bin and uploads to offset away from the bootloader when i first started.

repetier:
Good to know with the upload bug. I still need to add that method to server, but then I can add a test if it is still there after restart and delete it. The SKR 1.3 for example renames it to .cur like you also think it should happen.

PR #978, opened 2020-09-27
printTemperatures remove unused showRaw parameter

AbsoluteCatalyst:
compiling with LTO on the Due caused a weird compiler crash on this function. Looks like it's just because of the unused parameter.
(Did you still want a X show raw option?)


repetier:
At least showRaw is unused here. It was only used for debugging problems. But with the module structure I think we can omit it for now.

PR #981, opened 2020-09-27
Boot warnings add missing warning sounds

AbsoluteCatalyst:
Added the warning sound setStatus played before

### DIFF
--- a/src/Repetier/src/controller/gui.cpp
+++ b/src/Repetier/src/controller/gui.cpp
@@ -71,10 +71,10 @@ void GUI::update() {
             lastRefresh = HAL::timeInMilliseconds();
             if (HAL::startReason == BootReason::WATCHDOG_RESET) {
                 push(warningScreenP, (void*)PSTR("Reset by Watchdog!"), GUIPageType::STATUS);
-                // GUI::setStatusP(PSTR("Reset by Watchdog!"), GUIStatusLevel::WARNING);
+                Printer::playDefaultSound(DefaultSounds::WARNING);
             } else if (HAL::startReason == BootReason::BROWNOUT) {
                 push(warningScreenP, (void*)PSTR("Brownout reset!"), GUIPageType::STATUS);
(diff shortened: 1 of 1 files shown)


repetier:
BTW On my rumba32 with no usb power connection which is connected to a remote plug I always get a brown out warning after restart. This is correct I guess because when I unplug it the power goes down slowly so that is the definition of brown out, but it is not really nice to get this warning on every normal restart.

On the other side if I had a brown out during print I'd really like to know this to get the correct reason.

Do you have any idea on this problem except writing states to eeprom which will ruin them too quickly?

If not I'd rather remove the brown out message I guess. Otherwise I expect lots of questions like what is wrong - I only started the printer.

AbsoluteCatalyst:
Oof, that's not good. Most things I think of would die in sram unfortunately. A little complicated.
Doesn't the rescue system already write a special recovery byte to eeprom when a print begins though? 
(I haven't fully gone through the rescue system, sorry!) 
Could be used to check if we were printing and then display the error?
But then we'd need to decouple that in case they have the rescue system off etc.
  
I believe you're able to tweak BORRST on stm32f4's with different sensitivities too. I figure a real brownout would hang the processor for a while or behave erratically.

Alternatively you *could* also trigger an interrupt using the PVD peripheral as the power down occurs and write a small byte to the eeprom then and only then if printing, to reduce wear. Not sure if it'd make it in time though.

I just realized I made an error in updateStartReason for the STM's, looks like there can be multiple flags at once. I'll quickly write an update to use bitwise flags and fix that.

repetier:
I'm not so sure if this is a STM only problem. Often boards get still power from usb and if that fails it is quick. Or if you have a switch between power and board it is also quick. But if you unplug an otherwise unconnected printer it will get fading voltage causing brown out. And from server support I know there are quite some using switchable plugs to automate this.

So I think I will remove the display warning unless the rescue system was on in which case a brown out would be unexpected. But I will add a #define switch to force it as well just in case someone wants it every time.

repetier:
Ok, just committed the mod.

PR #982, opened 2020-09-27
Added Reboot - directAction menu option

AbsoluteCatalyst:
I've added it as the last menu option in mainMenu.


repetier:
Why do we need a reboot? Ok sometimes I also want to reset to reset the server connection and the printer had no reset key (no idea why some printers don't have them). Ok second thing is to stop a print from host.

And if we need it, it should be secured with a security question. That easy reachable a simple wrong click can otherwise ruin a print.

Also I think the entry should have a 
#ifndef PREVENT_REBOOT_ENTRY

protection. I now start also adding this for some parts a manufacturer might not want. In this example for example if he has a reset button which is much faster. For some eeprom entry just because he might not want users to change them with eeprom editor. E.g. a max pwm setting where higher values might destroy his hardware.

AbsoluteCatalyst:
Yeah I thought it'd be a good idea for a immediate reset in case of an emergency or the like.
I only have this new TFT with a reset button, didn't know they were common until recently. Removing the menu option if we have a button is a good idea. Could automatically remove it based on controller type.

I first had the option at the end of the configuration menu to hide it away and make it harder to click. I couldn't decide on a right menu though. What were your plans for the menu wizards? 

> Ok second thing is to stop a print from host.

I was planning on adding safe pause/stop/resume menu options via RequestStop: if we're printing from host, too. Think this might make reboot a bit redundant then?

> max pwm setting where higher values might destroy his hardware.

OH I've been meaning to suggest for the V2 docs: A small warning for the heaters & coolers page and HW PWM frequencies. It's possible to overheat destroy your board mosfets by keeping them in the linear region and switching them at a high frequency. Users should avoid PWMing their bed mosfets at more than >100hz as an example.

AbsoluteCatalyst:
Closed as it's not so useful anymore since we can stop via stop print. Almost done with a prototype immediate ramped down motion halt for pauses/stops too, should satisfy emergency pause needs.

repetier:
The immediate ramp down sounds very interesting. Would it also store not executed moves to be executed after solving the problem? I have a similar improvement in mind where we home xy when we loose a step detected by TMC and then continue print. That might use more or less same code also we can make an immediate stop then.

AbsoluteCatalyst:
At the moment I'm flushing the entire motion buffer after the ramp down, this is because I wanted to be able to control/move the extruder during a pause. I hadn't thought about storing them in a separate buffer, this is a good idea. I was only storing the line number sent from the host inside the motion buffer to then resend request from. (Isn't perfect atm since manual host moves get repeated)

For the losing steps problem, I stumbled upon a solution by accident; slowing down the motion3 timer frequency itself. This avoids changing any planned motions, and in *theory* i hope should avoid skipping steps even at high speeds. Whatever motion buffer id we land on, we use it's line number, step pos, etc to update our position to. 

How are you detecting a missed step using the TMC? Didn't know they could detect that. 

The only problem I was having is extracting the _exact_ step position it lands on. I'm landing in the middle of a motion3 block, so sometimes lastMotorPos is off by a little bit since we had some steps left.

repetier:
It is a quite complex problem together with host software. When they have send a line the forget about it more or less. If you allow any further command you can not use the resend trick and if you request pause you loose a lot of moves.

For the loosing steps problem a copy of motion1 not executed in a compressed way (only positions and feedrate as with G codes send plus start position) would be memorized and once we have position back through homing we push these moves back and unblock. So no host/sd card intervention would be needed.

The TMC drivers have stall detection in spreadCycle mode but not in silent mode. We program them to signal a failure on diag0 pin where we also can use them as end stop.  So all we do is enforce the louder spreadCycle and add some tests if we believe the signals. At lower speeds they become less reliable. That is also the same way sensorless homing works. See here:
https://www.trinamic.com/technology/motor-control-technology/stallguard-and-coolstep/

For fast pause also from server and be able to use printer further we need a cooperative mode to solve the the blocking problem. That also needs information from server that it would support this. Solution would be to add a CAP in firmware that it offers the function. Server would then indicate that it understood the capability and uses cooperative quick pause. Then you know we do not need to block host here.

Now there are 2 cases:
1. Pause coming from server. It will use a priority command to send that info and server will not send any "print" moves afterwards. It runs it pause script which can contain moves. There might also be a few moves in the input buffer which we also need to memorize.
2. Pause requested from firmware side. We signal the pause request to server (if supported) and start stopping and memorize left moves. We also memorize all moves until the pause started signal comes from server or buffer is full and we block.

Motion buffers get cleared and can be used for new moves and with a special gcode (or action in firmware) we move to last position start replay and continue with print.

That way you would have your pause and the stall detection with more or less same solution. I think with some clever state stages we can map that to same buffer. for n axis this should be doable with a buffer of (n+1)*4*(buffer size+6) byte. That is 760 byte for 4 axis and 32 buffer. Not sure if we can activate this on 8 bit boards later, but the 32 bit boards have enough memory for this.

Last problem as you already noted is  exact step position. Even if we know it, it would not really help. We only have PrinterType::transform which converts float position to steps but not vice versa. I think the solution is a step down flag in motion2 where we reduce speed with every block pushed to motion3 until we reach 0 speed. There we still know position in float position and make that start position and motion1 destination gets first buffered move.

What do you think?

AbsoluteCatalyst:
> For the loosing steps problem a copy of motion1 not executed in a compressed way (only positions and feedrate as with G codes send plus start position) would be memorized and once we have position back through homing we push these moves back and unblock. So no host/sd card intervention would be needed.

This is good, I wasn't sure how to handle SD yet as I never print from SD; so that'd knock out two birds with one stone.
 
> ...The TMC drivers have stall detection in spreadCycle mode but not in silent mode. We program them to signal a failure on diag0 pin where we also can use them as end stop...

Oh right! I totally forgot about that. The 2209's can do this in silent mode too. I think with the ramp down we should be safe at lower speeds.
 
> Pause coming from server. It will use a priority command to send that info and server will not send any "print" moves afterwards. It runs it pause script which can contain moves. There might also be a few moves in the input buffer which we also need to memorize.

Firmware will need to send another signal when it's fully finished stopping before the host runs the pause script I think.

> Last problem as you already noted is exact step position. Even if we know it, it would not really help. We only have PrinterType::transform which converts float position to steps but not vice versa...

I was curious about this; I was doing lastmotorpos/axisresolution, I figure we're losing some precision translating the steps back to float? 

> I think the solution is a step down flag in motion2 where we reduce speed with every block pushed to motion3 until we reach 0 speed. There we still know position in float position and make that start position and motion1 destination gets first buffered move.

This would be a good solution too, simpler than what I have now. How do we calculate how much to reduce the speed by each block? I was simply slowing down by time. (eg. 250ms to reach 0)

repetier:
> Firmware will need to send another signal when it's fully finished stopping before the host runs the pause script I think.
No it is the other way around. The unknown is the data server has already send before it sees you want this pause. So lets assume we use M124 which is marked as immediate stop.
M124 Sx sends states.
S1 => Server tells firmware I support this immediate stops
S2 => Server tells firmware I'm done with sending from g-code. Stop buffering.
S3 => Server tells firmware to immediately stop (needs to be parsed twice in this case. Once when we receive it as priority and later when it comes in order of execution to stop buffering).
S4 => send buffer and continue

Case firmware initiates stop seen from server log side:
send: N992 G1 X100
recv: // action: request_immediate_pause
Firmware goes already in blocking mode and buffers moves. New G0/G1/G2/g3 moves are buffered as well
Firmware gets N992 and buffers G1 that was already on it's way.
send: M124 S2
Firmware stops buffering, clears queue. New moves get executed again
.... Normal command exchange except server is paused
send: M124 S4
Firmware now executes buffer and server will add next lines from gcode.

Case server initiated pause
Server stops sending gcode but there are probably stored some in input buffer.
send: M124 S3
Firmware starts buffer as soon as it is parsed as priority command. Once it is received in order it stops buffering. Alternatively we could also require server to send M124 S2 afterwards. But that is just a point of possible failure when hosts implement it wrong.
.... Normal command exchange except server is paused
send: M124 S4
Firmware now executes buffer and server will add next lines from gcode.



> > Last problem as you already noted is exact step position. Even if we know it, it would not really help. We only have PrinterType::transform which converts float position to steps but not vice versa...
> 
> I was curious about this; I was doing lastmotorpos/axisresolution, I figure we're losing some precision translating the steps back to float?

You forget the non cartesian printer types. Especially the delta is very critical here. That is one reason why we also have a move steps commands that acts quite slowly. That is only used to undo steps we can calculate liek extra end stop offset in deltas or undoing z probes where we go to old step position to not loose any steps.

> > I think the solution is a step down flag in motion2 where we reduce speed with every block pushed to motion3 until we reach 0 speed. There we still know position in float position and make that start position and motion1 destination gets first buffered move.
> 
> This would be a good solution too, simpler than what I have now. How do we calculate how much to reduce the speed by each block? I was simply slowing down by time. (eg. 250ms to reach 0)

That is the real tricky part where this function gets so complicated. Once we stand still we  can take our time to fill a buffer. It gets even more tricky if we can not stop with the current move and need to continue with the next one. The question is how to get it to slow down quickly and how quickly do we do that. I think easiest solution is in Motion2 here:

```
        if ((m3->stepsRemaining = velocityProfile->stepsPerSegment) == 0) {
            if (m3->last) { // extreme case, normally never happens
                // m3->usedAxes = 0;
                m3->stepsRemaining = 1;
                // Need to add this move to handle finished state correctly
                Motion3::pushReserved();
            }
            return; // don't add empty moves
        }
```
We get the virtual step count for the next motion3 block. If we now have a flag for start blocking we could add after that something like:
```
if(quickPauseMode == QuickPauseMode::STOPPING) {
  quickSlowdown *= 1.05;
m3->stepsRemaining = static_cast<int32_t(static_cast<float>(m3->stepsRemaining)*quickSlowdown);
}
```

and more at the beginning when we would start a new block we test if quickSlowdown < 20 - if not we stop completely. Factors need to be checked. Factor 20 means a 200mm/s move would be executed as 10mm/s. We can compute the factor from the max allowed jerk compared to maximum possible speed. Each block is just 1ms if we have 1000 block per second . Factor 20 would be reached in 62 motion3 blocks which is very quick. The real factor instead of 1.05 could be estimated by max. acceleration *0.5 since we could decellerate during decelleration.

I know not cleanest way but I think it is very safe with minimal move reducing risk to stall and loose steps. So we also know position afterwards.

AbsoluteCatalyst:
Sorry; I did read this, and thank you for your thorough response. 

The explanation with M124 is great and helped me get an idea on how the state flow with hosts should work. (tangent: I'm wondering if we should use M124 or piggy back off of M539.)

>Alternatively we could also require server to send M124 S2 afterwards. But that is just a point of possible failure when hosts implement it wrong.

Right. Better to reduce the amount of commands needed from the host and stop buffering once M124 is parsed normally. I figure someone would make it a plugin of some sort for octoprint. 

> You forget the non cartesian printer types. Especially the delta is very critical here.

Aaah! I see. I was wondering about moveRelativeBySteps a while back... saw you slowed it down in one of your recent commits. 

> I know not cleanest way but I think it is very safe with minimal move reducing risk to stall and loose steps. So we also know position afterwards.

It's quite simple too, and looks promising. I'm curious to try implementing these, though I'm still slowly getting around to understanding how some things tie in together in the motion planner. Didn't think *increasing* the m3 steps would result in a slower motion.

repetier:
What does M539 do? Can't find a description for it. That was just a first example so not fixed.

Octoprint will surely make a plugin as soon as someone sees it. For repetier-server I make it of course.

>It's quite simple too, and looks promising. I'm curious to try implementing these, though I'm still slowly getting around to understanding how some things tie in together in the motion planner. Didn't think increasing the m3 steps would result in a slower motion.

Slow down yes. Each block is executed stpsRemainig times. So all we need to do is distribute the steps we want over more steps=time and we get a slow down even if we are officially accelerating. But this exponential slowdown will kill any move.

AbsoluteCatalyst:
```
void __attribute__((weak)) MCode_539(GCode* com) {
    if (com->hasS()) {
        Printer::setSupportStartStop(com->S != 0);
    }
    if (com->hasP()) {
        if (com->P) {
            Printer::setMenuMode(MENU_MODE_PAUSED, true); 
        } else {
            Printer::setMenuMode(MENU_MODE_PAUSED, false);
            GUI::resetMenu();
        }
    }
}
```
Quite similar to your M124 S1 Idea

PR #985, opened 2020-10-04
Fix emergency parser potentially wrong buffer size

AbsoluteCatalyst:
I had an issue where my SERIAL_BUFFER_SIZE on my due wasn't set to the default 128. This sometimes made emergency parser commands not work/get missed if the server was auto configured for the wrong size

This makes sure SERIAL_IN_BUFFER is set to SERIAL_BUFFER_SIZE if defined. Seems to fix the issue, though I'm slightly confused if it's meant to be its own independent size?

I also did this for SERIAL_RX_BUFFER_SIZE defined in the due's HAL in case it's used somewhere in the future.

### DIFF
--- a/src/Repetier/src/boards/due/HAL.h
+++ b/src/Repetier/src/boards/due/HAL.h
@@ -32,8 +32,13 @@
 
 // You can set different sizes if you want, but with binary mode it does not get faster
 #ifndef SERIAL_RX_BUFFER_SIZE
+#ifdef SERIAL_BUFFER_SIZE
+#define SERIAL_RX_BUFFER_SIZE SERIAL_BUFFER_SIZE
+#else
 #define SERIAL_RX_BUFFER_SIZE 128
 #endif
+#endif
+ 
 
 #ifndef HAL_H
 #define HAL_H
--- a/src/Repetier/src/communication/gcode.h
+++ b/src/Repetier/src/communication/gcode.h
@@ -35,7 +35,14 @@ enum class FirmwareState { NotBusy = 0,
 class SDCard;
 class Commands;
 class GCode;
+
+#ifndef SERIAL_IN_BUFFER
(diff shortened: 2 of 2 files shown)


repetier:
Not really sure why this should change anything. In my due RingBuffer.h I have 
#define SERIAL_BUFFER_SIZE 128
so same value as set.

But in general it is quite a problem that arduino has no official way to set buffer sizes for all variants working in all versions. It mainly depends on the variant used which defines the buffer size and maybe the arduino core for that platform like for due where it is the ringbuffer. Maybe I should check how Marlin does it that they can define the length for RX and TX individually.

AbsoluteCatalyst:
Mine is set to 256, I might have changed it at some point long long ago. The parser issues usually happened right before a print (while waiting to reach temp). 
Firmware would get a lot of start gcode commands at once filling up the parser buffer; but the server could keep sending commands since it thinks the buffer can support 255. Eventually missed commands, disconnect, etc.
At least that's what I think is happening. It doesn't help that arduino's due USB implementation is a bit of a mess...

Last I looked at marlin, they ported/hacked up the atmel ASF USB library for their due's usb serial and mass storage iirc. i assume they might just be modifying it directly

repetier:
Yes, when you modified the arduino ringbuffer you might get in trouble. I guess you did it there or do you know a way to inject different buffer sizes for due?

Yes, Marlin did some hacking. But I wonder it gets no problems or double definition using even more ram if the old implementation is there in parallel. Will also check in more detail.

AbsoluteCatalyst:
Just checked, doesn't look like there seems to be any sort of way outside of perhaps platformio pre-build scripts to change those macro definitions. Undefining the macros with .ini build flags isn't working either probably due to how far away the files are /includes/hierarchy/etc. 

(Also, I just realized there's actually a completely separate buffer used for usb. CDC_SERIAL_BUFFER_SIZE 512)

PR #986, opened 2020-10-05
executeFString gcode serial bug fix (multiple extruders)

AbsoluteCatalyst:
If the tool had a start/end script -> executeFString creates a fake gcode object (without defining a source) ->
Commands::executeGCode updates the global activesource to this nullptr.  
We tried accessing that nullptr here in prefetch. 
I think we weren't meant to check the activesource here, but rather the source this is called in?

Edit: Actually taking another look at this, should the executeFString code source be set to flashSource as well? 

Sorry, I totally stalked that forum [post](https://forum.repetier.com/discussion/8084/dev2-tool-extruder-gcode-fail) and already had my debugger out. lol.  

Edit2: There was another specific possible toolscript crash I found with minAxisEndstops[Z_AXIS] if you had PRE_RAISE and a probe with a different class name (even with ENDSTOP_NONE(endstopZMin) set)

### DIFF
--- a/src/Repetier/src/communication/gcode.cpp
+++ b/src/Repetier/src/communication/gcode.cpp
@@ -1270,7 +1270,7 @@ void SerialGCodeSource::prefetchContent() {
         } else // ASCII command
         {
             char ch = commandReceiving[commandsReceivingWritePosition - 1];
-            if (ch == 0 || ch == '\n' || ch == '\r' || !GCodeSource::activeSource->isOpen() /*|| (!commentDetected && ch == ':')*/) // complete line read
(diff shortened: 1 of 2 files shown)


repetier:
Took me a while seeing the problem. GCodeFlashSource is not used at all. And the execute command does not directly need it. Problem is when a external source like display fetches some data and activesource is null from the execute running. And G28 takes long so it is more likely to happen. I think your solution is perfect.

And no problem for hijacking - I like it. Currently busy finding an ugly bug in windows server and as you know I'm always quite busy. So any help is welcome.

PR #994, opened 2020-10-18
Baudrate config menu - update baudrate when exiting menu

AbsoluteCatalyst:
I noticed changing the baudrate in the config menu didn't seem to actually change it when selected. 
I've changed it to update immediately once we exit the menu. 

To do this I've added serial end commands in each of the HAL's serialSetBaudrate to restart the serials at the new baudrate.
There's an included small check to avoid resetting the USB CDC connection. 
(I didn't add this check to the SAMD as I saw your previous commit where you excluded it from the gcodesource check)

It appears we also have a small bug where the baudrate currently gets reset to BAUDRATE on reboot. 
I *think* this is only happening with SDCard eeprom. I'll double check that when I get back.

### DIFF
--- a/src/Repetier/src/boards/SAMD51/HAL.h
+++ b/src/Repetier/src/boards/SAMD51/HAL.h
@@ -614,8 +614,10 @@ public:
     static inline void serialSetBaudrate(long baud) {
         // Serial.setInterruptPriority(1);
 #if defined(BLUETOOTH_SERIAL) && BLUETOOTH_SERIAL > 0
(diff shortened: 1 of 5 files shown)


repetier:
Ok, think it is better do update directly. Especially if you reconnect without reset it would be unexpected otherwise.

AbsoluteCatalyst:
Can't reproduce that wrong baudrate on reboot bug anymore. Must have been something on my side back then, nevermind.

PR #997, opened 2020-10-19
Add P<period ms> parameter to M155 autoreport

AbsoluteCatalyst:
Quickly added a P<milliseconds> parameter to let us change the rate we get auto temperature reports (for debugging etc). 
Minimum is 100ms, max is 10 seconds. 
Will always safely reset back to 1 second if P is omitted for best compatibility/host reconnects/errors etc.


repetier:
Might be handy if you need high resolution to see when temperature rises in sub second area.

AbsoluteCatalyst:
Quite useful with repetier host's large temperature graphs, helped me visualize some sporadic wiring issues and adc problems.

PR #1001, opened 2020-10-19
Fix extra cold extrusion warnings on moves

AbsoluteCatalyst:
Added a quick check to make sure we're moving the E axis before displaying the new cold extrusion warning.

Looks like we have a few places at the moment where we do something like oldCoordinates[E_AXIS] = currentPosition[E_AXIS] rather than IGNORE_COORDINATE before the moveByOfficial/Printer calls

Alternatively we could update those places to use IGNORE_COORDINATE i think.
(Wasn't sure if they may have been intentional and didn't want to break them.)

Edit: Actually, it looks like there's not really that many places that happens. Mostly in ZProbe.
Want me to just fix the responsible moveByX calls without IGNORE_COORDINATE on the E_AXIS and update this pull instead?


repetier:
Makes sense to test that too. Sometimes you just cache old positions to go back to them later and then it is same extruder position. So this catches the problem and reduces warnings.

PR #1002, opened 2020-10-19
Fix SKR e3 mini build

alexsomesan:


alexsomesan:
mistakenly opened against master. Will re-open

PR #1006, opened 2020-10-20
Fix STM32F1 compiling - missing setInterruptPriority

AbsoluteCatalyst:
Ran into this too while setting up my HW TMC's. 
STM32Duino has setInterruptPriority for software serial and timers, but not HW serial. Didn't know that.

Maybe I shouldn't have added it here since it seems only the Due has that function?


repetier:
Setting priority is in general a good idea. It should be high to not loos anything but lower the stepper isr so we get best stepper timings possible. So hope this is still the case. After all it will have a priority, just you can't set it.

AbsoluteCatalyst:
Oh yeah it still helps of course. 
I meant that more as like; I would have tried making a more universal way to update them, had I known. 

Found a UART_IRQ_PRIO macro config for the STM32's in the meantime though

PR #1010, opened 2020-10-25
Allow stop/pause/continue host prints from menu

AbsoluteCatalyst:
I modified the SD stop/pause/continue directactions to use the Printer::stop/pause/continuePrint functions since they automatically handle if we're sd printing or host printing and run the right command.

Good in case we're not around our pc/phone at the time to control the host.


repetier:
Good idea - unifying makes things more simple:-)

PR #1012, opened 2020-10-26
Allow importing smaller grid size matrixes, add P<gridsize> parameter

AbsoluteCatalyst:
Think it'd be a good idea in case someone wanted a different grid size at runtime, or wanted to probe slightly faster. (I saw that one topic on the forums.)

GRID_SIZE is now MAX_GRID_SIZE to represent a maximum grid size allocated in memory. Grids smaller than this can now be imported and run. Configurations have been updated to reflect this. 

G32 now has a new parameter P<gridsize> which can define how many grid points to measure at run time. Min = 3, Max/omitted = MAX_GRID_SIZE

PS. Should I be updating the version parameter on EEPROM::reserve whenever we do any eeprom additions?


repetier:
Good idea.

The checksum contains size and version. So if either of them changes it will reinit eeprom anyway. So no need to increase version in this case.

PR #1014, opened 2020-10-27
Only kickstart if new PWM is higher than current PWM

AbsoluteCatalyst:
Small fix, was attempting kickstarting even at 80-100% speed.


repetier:
Makes sense to assume it was already running so not needed.Would only fail if min speed is too low and that is a configuration error:-)

PR #1015, opened 2020-10-27
Host priority mcode/button controls

AbsoluteCatalyst:
You reminded me, I was going to show you this small concept with prioritized mcode for host controls (GUI buttons, etc) that's separated from normal mcode. I'm running this by manually editing my XML file atm.

It works by shifting two flag bits to the end of mcode commands, only parsed by the emergency parser when they appear. 
Avoids adding any new parameter (may be in use), avoids messing with timed mcode inside gcode files, and avoids any other mcode numbers we'd probably ever use. I've added a new cap sent to the host: HOST_PRIORITY_CONTROLS, so no conflict. 

Think this is really good for eg. repetier-server's fan drop-down control, especially if my slicer failed to detect it needs to up the fan during long moves on thin objects. 

Fan drop down would now send eg M49258 P0 S200. Toggling the case light sends M49256 S0 even if warming up etc. 

It's not *so* useful for the temp controls, neither for flow/speed (still motion buffered). But might be useful for CNC/laser intensity/syringe tools I think, and anything else. 

(Actually it's useful for my chamber heater since it ramps up it's fans immediately, later servo vents too.)

What do you think?


repetier:
I like the princible idea but the solution is not versatile and easy enough for users. I can't add known mcode to 49152 so fast so that is a unfriedly solution to mark them. Much easier is when we assume all M codes are < 10000 then we could say add 10000 to it to run it immediately.

Also now we have a way to mark them for immediate execution lets make ALL mcodes executeable immediately. With the solution we have that would be no problem (marlin will not be able to this with the fixed parser for emergency but not our problem here). We have a mcode execution function so if we see a immediate command we:
mark it immediate with a flag
subtract 10000 form mcode
run the execute mcode
now we need to prevent double execution. One way would be change M code to a dummy already executed command so on regular call it will do nothing.

With the flag we can still do special handling like with fan flag.

In server I add that when cap is present all M codes 10000-19999 are also priority commands.

More gain and easier to memorize for users. 

What do you think?

AbsoluteCatalyst:
That's a good alternative too. I originally slightly intended it to be moreso as a hidden command to avoid users accidentally breaking something incase there were riskier commands. But you're right it would be good if they could. 
So you're saying, 
1. 10xxx priority MCode arrives at testEmergency
2. subtract 10000 from it and turn on a "isPriorityTypeX" flag inside it's GCode class
3. run processMCode immediately inside testEmergency with the new subtracted mcode
4. unset M after that call
?

AbsoluteCatalyst:
Whoops, there was actually a slight error in this pull. 
```
    } else if (com->hasM()) {
        if (!com->isImmediateM()) {
            processMCode(com);
        }
    } else if (com->hasT()) { // Process T code
```
should be 
```
        if (!com->isPriorityM()) {
```
(This is the way i'm avoiding running the command twice right now.)

AbsoluteCatalyst:
Errr. Wait what? I'm sorry, there's actually a bunch of things wrong with these commits in this pull! I didn't properly pick and apply updated commits from my main branch. One sec!

AbsoluteCatalyst:
Okay. Take a look at that last push. I applied your suggestions too. Quickly tested them and they work with the server. 
Even though i'm doing:
```
gcode.M = gcode.getPriorityM();
Commands::processMCode(&gcode);
```
That same gcode object doesn't get reused afterwards in executeGCode, so it appears safe to change it's M value without repeating it.
Tested it out on the server and I can send any M code immediately now.

repetier:
Yes I think that solution should work great. Will add server support as soon as I can. Just want to finish the automatic delta geometry calibration. It's nearly finished:-)

AbsoluteCatalyst:
Oooh, awesome. And thanks for the server support!

repetier:
No problem. Just a few lines and of course must our products support our firmware as good as possible:-)

PR #1017, opened 2020-11-02
PWM Kickstart - only from 0 & parameter names fix

AbsoluteCatalyst:
karl_ranseier suggested that it might be better to just kickstart only from zero pwm.

Figure since the fans are already moving with some sort of inertia, they don't really need any more kickstarting afterwards, even if they're spinning slowly. 
Seems to work okay with all my fans so far. 

IO_PWM_KICKSTART's parameter names needed a slight fix on some macros.


repetier:
Yes, makes sense.

PR #1020, opened 2020-11-03
Attachable/Detachable HW-Endstop interrupts when needed.

AbsoluteCatalyst:
Smallish add-on to the endstop drivers. ( It should also work with merged endstops as well. ) and a function to Motion1 since it actually works well complimenting endstopForAxisDir. 

I've changed homing to attach the specific endstop we're going to use per axis. So far so good. 
ZProbe only attaches the ZProbe endstop at activate() and detaches at deactivate etc. 

Motion1::setHardwareEndstopsAttached(bool attach, EndstopDriver* specific = nullptr)
Attempts to detach/attach all axis HW endstops unless given a specific endstop driver to which to turn off.
Won't turn off any endstops if alwaysCheckEndstops is enabled. 

Hope I haven't missed anything with the other machine types though.

### DIFF
--- a/src/Repetier/src/drivers/endstops.h
+++ b/src/Repetier/src/drivers/endstops.h
@@ -16,6 +16,10 @@
 
 */
 
+#ifndef ALWAYS_CHECK_ENDSTOPS
+#define ALWAYS_CHECK_ENDSTOPS 0
+#endif
+
 class EndstopDriver {
 public:
     // Called by stepper driver before each step to update state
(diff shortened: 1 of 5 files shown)


repetier:
Looks good so far. For delta and dual z end stop these are motor end stops, but they get covered with the global z axis that merge them for global z axis. So I think for the current state it covers all.

The pure motor end stops for stall detection might get a special flag or different type when we do not want them to be disabled or work as stall detection. That way we can keep this without problems I think.

AbsoluteCatalyst:
That's a good idea, stall detection slipped my mind. Maybe something in the drivers like; isDetectingStalls()/setDetectingStalls(bool). 
Then the steppers classes can set the flag to their private endstops whenever necessary to be skipped by the motion1 function

PR #1021, opened 2020-11-03
Fix SKR Mini - BTT TFT Simulation mode display

AbsoluteCatalyst:
Figured out why I couldn't get the little simulation mode to render. TFT appears to only accept an active low clock signal.
Changed the SPI mode only for STM32F1's inside u8g2's st7920 display info.


repetier:
The st7920 is in deed very picky on timings. Took me a while to get timings that worked with the due.

AbsoluteCatalyst:
Definitely. Hopefully the timings needed are all the same across other users with it.

PR #1023, opened 2020-11-04
Add AUTOREPORT_SD_STATUS cap & M27 C0 S<sec>P<millis> SD status Autoreports

AbsoluteCatalyst:
Just a quick clone of autoreport temperatures. Only autoreports byte positions during a SD print. Requires SDSUPPORT. 

I've brought in the S<seconds> parameter to M27 and M155 as well for better compatibility across hosts. 

Setting either P or S to 0 will disable autoreporting. Using both will just combine them.

Omitting P _and_ S for M27 will just print out the status as normal. Adding a C parameter will also print out the filename.


repetier:
Nice addition when we start using sd print in server.

AbsoluteCatalyst:
Looking forward to any SDCard features/tools for server.

Speaking of server, do you happen to share the unminified/uncompressed .js source files for the main web interface anywhere? eg. app-min.js?

PR #1025, opened 2020-11-11
Implements G10/G11 Firmware retractions/config menus/autoretract

AbsoluteCatalyst:
I've used the setting names/macros we had left planned for FW retractions, and now "Retractions" should appear under Configuration in the GUI main menu. 

(Did you want specific retraction settings per extruder tool instead?) 

~~I haven't noticed any missteps or issues, but I could be wrong on other printer types.~~
~~Doing more test prints now in case I missed something.~~

Edit: It was very broken.

- Added missing AUTORETRACT_ENABLED for default configs.
- ZHops lead to wrong positions if Z changed between hops. Now properly using Z ToolOffset's.
- ZHops now respect moveFeedrate, and print move accelerations. (secondaryMove)
- Added the missing feedrate & flow multipliers to speed and extrusion amount.
- Fixed some menu titles & increments.
- Retractions were inconsistent. Switched to printer coordinates and they've been corrected. I have an implementation using steps instead, if you think I'm still going about it the wrong way.


repetier:
Looks very good. We are coming closer to a feature complete firmware.

I'm currently adding a new feature to allow out of order commands even when communication is blocked. More or less like emergency commands but they get removed to free the buffer so we can add next command when host supports this. Allows for example many babysteps without delay from server interface. Comes also along with host prompt support added to firmware and server so we can have dialogs on server side as well if required. Already added warning/error pop ups there as well.

AbsoluteCatalyst:
Ah that sounds great. Good to fully remove the buffer restriction for those commands. Should speed up a lot of host controls. Is this going to be a new version/update to the binary protocol? 

Eager to see the host prompt support too. 
I think I remember there was a warning prompt of some sort on the server years ago when printWarnings appeared.

I was working on something related with comms. I was quickly writing a small experiment for GCode quote text parsing, so we could send MCode with parameters and text at the same time.

repetier:
Not a change to binary protocol. If ooo is enabled a command send and detected will not send "ok 2344" but "ooo M108" and remove it from input queue, so we can read again more without front command being removed. Works only for small commands fitting into remaining buffer and also only for commands reported to use ooo.

Mcodes with text and parameter might be a problem with binary protocol resp. the converter. Protocol can store it I think, but server/host must detect what is text and what is parameter. Need to think about that. Maybe add some [X] marker or so in server text command list. or put anything starting " " as text or content with 2 letters in a row.

PR #1026, opened 2020-11-11
Fix M202 Max Accel/Travel Accel

AbsoluteCatalyst:
M202 was using the wrong variable. Updated the command list comments.

Edit: Added isSecondaryMove() for the extruder tool. Will return true if it's an extrusion move OR a G1 move.


repetier:
See the error.

PR #1029, opened 2020-11-12
Menus - round to multiples of increments

AbsoluteCatalyst:
Only rounded upon leaving a min/max limit or changing sign so that fine tuned increments stay incremented.


repetier:
You can not use floorf/ceilf/roundf on long integers! That would limit values to 8388608 due to float conversion. Not sure if that is important anywhere but better be safe then get problems from strange roundings. Can you update the long version?

AbsoluteCatalyst:
Oh crap whoops. Give me a sec i'll fix it.

AbsoluteCatalyst:
Yeah that should be all. Just wanted to make sure it worked with negatives properly.

PR #1030, opened 2020-11-13
Due - Fix Freeze/Hang on USB disconnects.

AbsoluteCatalyst:
Fixes a really old problem with the native ports on Arduino Due's and the Arduino Library. 
Disconnecting the native port and then trying to serial write will cause an infinite loop. 

This works around the issue by checking if the USB port is in a suspend state before trying to write bytes.

Learned about using the suspend state from STM32. Used to check if they're connected or not in their HAL. 

(I think OS's have the choice to implement/broadcast suspend functionality or not, but so far it reliably changes state from my Pi and Windows desktop)

### DIFF
--- a/src/Repetier/src/communication/gcode.cpp
+++ b/src/Repetier/src/communication/gcode.cpp
@@ -1238,6 +1238,14 @@ int SerialGCodeSource::readByte() {
 #endif
 }
 void SerialGCodeSource::writeByte(uint8_t byte) {
+#if defined(DUE_BOARD)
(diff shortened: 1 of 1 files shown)


repetier:
Never registered that problem. Maybe because I'm using mainly serial connector and only with server:-)

AbsoluteCatalyst:
I usually only use server, but always though native. 

I think other users using native might have thought it was normal for the due to reset (because watchdog) like the serial port does too.

PR #1032, opened 2020-11-17
Implements M575 - Change Serial baudrate

AbsoluteCatalyst:
Added M575 to change baudrates. Matches the other firmwares and our menu option. M575 B<baudrate>
Only 38400, 56000, 57600, 76800, 115200, 128000, 230400, 250000, 256000, 460800, 500000 are accepted.
Can't change per serial yet.


repetier:
Great. Have already added store to eeprom and in G4 we can now use S and P in parallel e.g. G4 S7 P80 for 7080ms.

PR #1036, opened 2020-11-18
Move SAM/AVR to C++14

AbsoluteCatalyst:
I saw STM32's use 14' by default but the SAM's and AVR don't. 
This should make them equal and gives us a few nifty things like constexpr functions with if's/variables, auto return types etc.


repetier:
Good idea. Not that we become problems from using C++ 14 somewhere otherwise:-)

PR #1037, opened 2020-11-19
HW IO_Endstop - Lambda callbacks

AbsoluteCatalyst:
Small update to HW endstop switches, now uses a lambda as a callback passed through the constructor. Slims the structure a bit. Removed IO_TARGET_INIT as it's no longer required. 

I created a new global typedef; void_fn_t, in constants.h. Easier use of simple functors in the future.


repetier:
Good use of lambda. Still not used to using them in C++ also I use them every day in javascript.

PR #1041, opened 2020-11-20
Update u8g2 Library -  From v2.24.3 --> v2.27.6

AbsoluteCatalyst:
I've moved it to within the Controller directory instead of root since it seems more fitting.

I merged any Repetier specific changes I found, in. But there weren't many. 

Most Repetier changes were inside of U8x8lib.cpp (The STM32F1/due speedups), some others in u8g2.h (settings), and one change in u8x8_d_st7920.c for due boards.

Did I miss anything? Working well so far on this Mini'.


repetier:
What I remember are the st7920 changes. STM changes are from you I think. Will see if we get other problems. Will later test st7920 on stm32f4. But at least I know then where to look for. In general it is good to be up to date again.

AbsoluteCatalyst:
Yeah the atm was from me, just for the stm32f1s though. 

Sdfat is next. He just released version 2, 11 days ago.

repetier:
Good to know you are on the sdfat part. Just saw 2.0 supports now extFat format as well.

My next things will be new temperature controller with pre learned pwm/temp curve for fast settling and low overswings.
I'd also like to play with input shaping but have some problems with the math and many articles are too abstract. I need to know how to convert the impulses to accelerations.

PR #1042, opened 2020-11-20
Feature - ZProbe Progress Screen

AbsoluteCatalyst:
Adds a small animated leveling progress screen for both 20x4's and u8g2's. 
This is primarily for grid based leveling since it can take a fair bit of time.
But I believe it could be fit to work with the other leveling methods too.

(Sorry about the u8g2's probeProgress function being somewhat visually bloated, used a fair bit of constexpr margin variables to handle things.)

It looks like this:
image


repetier:
Looks really good and for those having many probe points it is in deed a help I think.

PR #1043, opened 2020-11-20
HAL - Add handlePeriodical

AbsoluteCatalyst:
We'll definitely need this in the future for HAL specific tasks I feel. 
It's empty for all the HAL's right now though.


repetier:
Not so sure if we really need it since every module can also have a callback in periodical, but you never know and it won't hurt.

PR #1044, opened 2020-11-26
ZProbe - G32 add R<repetitions> A<useMedian>

AbsoluteCatalyst:
Hi. I've added these two parameters to let us change the repetitions and use-median setting for probe runs at runtime as well. Works across all leveling types. 
Z_PROBE_REPETITIONS and Z_PROBE_USE_MEDIAN are now the default settings if the parameters are omitted.


repetier:
Great idea to give it more flexibility.

PR #1045, opened 2020-11-27
Add Babystep Z control to Tune menu

AbsoluteCatalyst:
For mid-print changes.

Related: You think it'd be worthwhile to print the BabystepZ: messages in the periodical timer, and only after an amount of time (3ms-5ms) & if changed? (only do this if changed from the controller)

STM32F1 takes 500-700us per message...
My due takes 80-100us~
Maybe unnecessary...


repetier:
I don't think a change is required. It writes any change from controller or M290 since all call the PrinterType::M290 to change, so hosts interested can listen. You can also send M290 with Z0 to query value if wanted, so no need to pollute output with periodical messages in addition.

AbsoluteCatalyst:
Sorry, I explained it poorly. The idea was that:
Scrolling baby steps very quickly intentionally *wouldn't* send the message until it settles down, and just once.
Like a sort of throttling of the prints.
I remember you were trying to speed it up recently. 
My theory was it would help speed up baby stepping and reduce any jitter to the motion system. 

But in the end I don't think the message takes up enough time anyways, and I also found some optimizations to speed up prints quite a bit.

repetier:
I also don't think it is worth the extra memory. Reporting is quite fast and adding it step by step is good. Big steps get throttled by the maximum babysteps per motion3 block anyway.

PR #1047, opened 2020-11-27
Menus - Add Extr. Fine steps/mm menu & Misc Fixes

AbsoluteCatalyst:
- Added Coarse/Fine menus for the extruder resolution menu to match the axis controls.
- The target temp increment feels far too slow on encoder speed 2, i've added a small check to increase it to 5 when on speed 2.
    This is the only control so far that I felt needed this though, everything else seem fine.

- A few incorrect unit display fixes (MaxPWM, Extr. offsets, etc)

### DIFF
--- a/src/Repetier/src/drivers/heatManager.cpp
+++ b/src/Repetier/src/drivers/heatManager.cpp
@@ -10,7 +10,7 @@ void menuSetTemperature(GUIAction action, void* data) {
     HeatManager* hm = reinterpret_cast<HeatManager*>(data);
     float value = hm->getTargetTemperature();
     DRAW_FLOAT_P(PSTR("Target Temperature:"), Com::tUnitDegCelsius, value, 0);
-    if (GUI::handleFloatValueAction(action, value, hm->getMinTemperature(), hm->getMaxTemperature(), 1)) {
+    if (GUI::handleFloatValueAction(action, value, hm->getMinTemperature(), hm->getMaxTemperature(), (ENCODER_SPEED == 2) ? 5 : 1)) {
         hm->setTargetTemperature(value);
     }
 #endif
@@ -32,7 +32,7 @@ void menuHMMaxPWM(GUIAction action, void* data) {
 #if FEATURE_CONTROLLER != NO_CONTROLLER
     HeatManager* hm = reinterpret_cast<HeatManager*>(data);
     float value = hm->getMaxPWM();
-    DRAW_FLOAT_P(PSTR("Max. PWM:"), Com::tUnitDegCelsius, value, 0);
+    DRAW_FLOAT_P(PSTR("Max. PWM:"), Com::tUnitPWM, value, 0);
     if (GUI::handleFloatValueAction(action, value, 1, 255, 1)) {
         hm->setMaxPWM(static_cast<uint8_t>(value));
     }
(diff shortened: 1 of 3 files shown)


repetier:
Not sure about the encode speed change. How do you then change to a full degree if you use 5 degree steps?
Also it is not slower of faster than with any other encoding. Encoding is selected so that one click is one step or should be set that way.
In V1 we had a speed dependent increment which might a better solution. If you change with higher speed we take x2 or x4 times the step size. Could be done storing last action and time. And if we have same action we have a repeat time to use. Maybe add an extra parameter to disable or slow down for more sensitive changes.
What do you think?

AbsoluteCatalyst:
I was wondering about speed dependant increments, yeah that'd work too.
I set it to 5c increments as I figure 1c changes aren't so common. It'd always land at multiples of 5, which seemed ok. 

I haven't looked too deep into how the encoder pulse train works when it comes to the encoder speed. 
i can set my display to setting 1 and I'll have single steps too, but if I start spinning faster, it'll start skipping menu options and be too fast and uncontrollable.

repetier:
When it is hard to hit menu positions your encoder speed is too fast. Hard to hit a menu between 2 encoder clicks:-)
But 1°C steps are required. While I don't see a problem here I know enough users wanting exact temperatures. Some even would like 0,1°C steps but that is too much also since we use integer for communication. So I will revert that back and we later can add speed dependent increments.

AbsoluteCatalyst:
Alright that seems fine. I think I may have focused a bit on the hotend while forgetting we have other heater types that may have more precision. 

Got any idea how the speed dependant ramp up equation should be? Just simple linear? or was there a bit of a exponential type curve to it?

repetier:
Exponential is normally bad. Gets much too fast high. Should also be capped, hence the extra parameter I mentioned. 1 would be just increments so it stays like it is. Otherwise it would be the maximum step size I think.

I think linear would be best solution with a maximum time of 500ms (lets make it a macro value to adjust). Anything faster then 2 clicks per second would be increased until I think around 100ms.
so as algorithm we would get something like:
if(repeatTime > MAX_REPEAT_TIME) {
 single step
} else {
if(repeatTime < MIN_REPEAT_TIME) {
  repeatTime = MIN_REPEAT_TIME;
}
maxSteps * (MAX_REPEAT_TIME - MIN_REPEAT_TIME)/repeatTime;
}
Timings might need some testing to find most harmonic values, but that is what I think would work best.

What do you think?

AbsoluteCatalyst:
Yup, think that'd be easier on the user to configure too. 
I can try that snippet later today.
Eeprom/menu configurable would be nice too.
(Think we still have plenty more room right?)

repetier:
Thanks for adding it. I'm just in the end of server update. Have updated config wizard considerably but need lots of tests with all the firmware to verify it works.

Yes, eeprom would be nice since users have different preferences. With eeprom we have 4kb for all data we store. Plenty depends on number of modules with eeprom usage I'd say. But what takes most is grid for bump correction so guess we have no real problems. And for most systems we could theoretically increase the allowed eeprom. For stm we only use flash anyway and radds has bigger eeprom then 4kb i think. 4kb is just the smallest from AVR boards.

AbsoluteCatalyst:
Great. I think the e3 mini v2's eeprom was 4k too since it uses a i2c chip. 

Also, this update has the new prompt and out of order support you worked on recently? 

I was planning to help out with jam system (I haven't set up the mini's jam config example) once I finished the new SdFat V2 updates. Almost done with that, just redid some small things.

repetier:
Yes, the out of order is already implemented. If you want the linux version of server 0.94.4 are up to date for testing. Just change download link and test it. Prompt support is also working now and uses out of order commands for V2.

PR #1048, opened 2020-12-03
Fix missing bed/chamber tune controls

AbsoluteCatalyst:
Just noticed controlling the temps on these two were missing while printing. 

I think it may also be a good idea to remove the home/move options from the tune menu? 
Or hide them unless paused, maybe.


repetier:
Agreed. During print homing and move make no sense. Just like any other command requiring moves.

PR #1050, opened 2020-12-12
SdFat V2/exFat support/various SD improvements

AbsoluteCatalyst:
Hey. Finally got some more free time and finished this. I did take some small liberties and added a few things, but *should* all still run the same. 

- I didn't replace the library's print functions this time around. We didn't really use them, and I think it should make it easier to update the library in the future especially since there's more to come soon. (They _do_ accept arduino-based print classes now though)
- I did make a small text file (in it's src) with any custom changes to the library written down.

- exFat support is on by default for ARM. It can be disabled for flash savings via setting SDFAT_FILE_TYPE. (1 for FAT16/FAT32, 2 for exFAT, 3 for FAT16/FAT32 and exFAT.) But shouldn't be necessary on any of our platforms.

I slightly redid our sdcard class:
- sd.sdactive and sd.sdmode have been replaced with a SDState enum. (::SD_PRINTING etc)
- sd.file is now sd.selectedFile (only meant for writing/printing) and sd.fat is sd.fileSystem
- Two new typedef's, sd_file_t (for files) & sd_fsys_t (for file system). Handy if there's ever a library change (HAL's, FatFs, Mass Storage, etc).
- Changed some function names for clarity.

- Redid the automount logic: Now settles for 250ms (async) before mounting. 
- - Will quietly retry a mount init 3 times before declaring it a failure. (SDState::SD_HAS_ERROR)
- - Displays a helper error message for a common SDSS pin config error.
- - Detect-pin level is also reported in the error.
- - A mount resets eprSyncTime in case it was already scheduled.
- - Unmounting (either auto or manual) during a print now stops the print and shows an error message.

- Added helper doForDirectory template, runs any kind of function/lambda on every item in a directory. (recursive-optional)
- Added helper validGCodeExtension, accepts .gcode, .gco, .gc, and .g extensions.
- Added getCardInfo for getting volume labels, volume size, usage bytes, file count, folder count.
- - The user's sdcard volume label is shown in the menus. Looks cool & will be good for multiple SD's later. 
- - Small stats shown after a succesful mount (sd.printCardStats()): 
13:59:13.520 : Label: mylabel | exFAT SDXC | Volume Size: 63.83 GB | Usage: 6.64 GB (11 files, 0 folders found.)
- ls can accept a directory. default is root.

- Reduced the verbosity/print usage of bumpmap importing and exporting. Uses more common/standard reports.

- SD writing works fine (35Kb/s~ over serial). So does printing. 
- SD writing now has a 5 minute inactivity timeout.
- Now PreAllocating 8kB before M28 writing to files. (2kB on AVR) This may not be necessary, but could reduce wear.
- I went over JSON outputs too. They're working as well (the info functions have outdated search strings though.)

Tell me if I missed anything or removed/changed something I shouldn't have. I think everything is in working order.


repetier:
Wow, what a big commit. So hard to say if something was forgotten. I like the addition of the enums. Last days did the same on my server code for some parts to make it more easy to read and prevent illegal values.

Also the settling when inserted is a good idea I think. So in total a great job and only testing and time will tell if something is missing.

AbsoluteCatalyst:
Thanks.
And yeah, I had some trouble remembering the sd.mode/active numbers and what they meant sometimes. Enum states help.

I had some worries when I moved around the prints that the hosts look for (eg. sd removed/inserted.) but I corrected this and they seem to be in order again. 

There seems to be some very minor bugs with exFat cards right now. eg. It seems we need to use .sync(); even if O_SYNC is enabled. Not really a problem, and still good to be ahead of the curve with exFat. Very few 32gb sd cards on amazon now. In a year people might not even have a choice.

repetier:
In deed. Didn't realize exFat is something different then fat until someone told me he could not mount exFat usb sticks on server image. Was easy to add support but I needed to add the package. On next debian it might already be a default package.

PR #1051, opened 2020-12-13
GUI - SD Menu scroll bar

AbsoluteCatalyst:
Forgot to upload this tiny gui addition to help with large folders. Only shows up for U8G2's.
A thin simplish scrollbar appears at the right of the screen while the rows change. 
It disappears quickly after we stop moving to avoid blocking any important text. 

Made for the SD menus, but coincidentally works for all/any menus as well. I only added them to the sd menu. 

GUI::showScrollbar(GUIAction& action) <- auto calculates the bar based on top row, current row, length etc. 
Just drop before a menuEnd and the bar should appear if there's enough items on the screen.

GUI::showScrollbar(GUIAction& action, float percent, uint16_t min, uint16_t max) <- allows manual control

### DIFF
--- a/src/Repetier/src/controller/drivers/Display20x4.cpp
+++ b/src/Repetier/src/controller/drivers/Display20x4.cpp
@@ -1208,6 +1208,11 @@ void GUI::showValue(char* text, PGM_P unit, char* value) {
     printRowCentered(3, GUI::buf);
 }
 
+// No scrollbars for the 20x4's 
+void GUI::showScrollbar(GUIAction& action) {
+}
+void GUI::showScrollbar(GUIAction& action, float percent, uint16_t min, uint16_t max) {
+}
 //extern void __attribute__((weak)) startScreen(GUIAction action, void* data);
 //extern void __attribute__((weak)) printProgress(GUIAction action, void* data);
 // extern void __attribute__((weak)) mainMenu(GUIAction action, void* data);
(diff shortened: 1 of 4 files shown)


repetier:
Cool idea. Need to check with config and tune menus how it looks. I think it could be a nice addition to see where in the menu you are.

AbsoluteCatalyst:
Just came back in and tested it out now on all the menus and it seems to mesh fairly decent system wide. Yeah try it out when you get a chance. Menus such as extruder config are quite long. 

( We haven't done the TMC drivers settings menu yet right? I could have sworn there was one. Weird. )

repetier:
I will. In that case we can put it directly in menuEnd.

Checked it and also can't see any TMC menu also thought we hat them. Did one of us remove them with a merge involuntary? Haven't removed them on purpose in any case.

AbsoluteCatalyst:
Ah, It looks like we just haven't gotten around to them. There's menus for the adjustable resolution drivers, but besides that git doesn't show anything.

repetier:
We also have it for mirror so all get called. So maybe that is what we were thinking about. And it is in eeprom. Question is also how much sense does it make in menu. Nothing you would change constantly. Maybe current since some errors can change current settings.

AbsoluteCatalyst:
Current was the first I thought about too. Only thing I ever change at any frequency. 
Stealth?

repetier:
Not sure about stealth. With hybrid and homing it changes automatically as well. Also if I think about prusa there you can set it for the print so you decide loud or fast for all motors. Actually I like the hybrid thing so your speed just decides and if you want crash detection you have no choice anyway.

PR #1052, opened 2020-12-13
Steppers - Fix STEPPER_OBSERVEABLE TMC driver eeprom

AbsoluteCatalyst:
Are we still using this module for jam detection? I remember it was in the default IO configuration. Looks like the samplesystems changed a bit from last I looked.

Stepper Observable has a small bug where it'd prevent the tmc eeprom from showing if we added it's class to the motor list instead of the driver itself. 

This adds the driver's missing eepromHandler call, and adds a parent pointer check in motorIndex to find the correct motor list index.


repetier:
Yes, it is still used for jam detection. Only way until all have a counter.

PR #1053, opened 2020-12-14
GUI - waitScreen newline & Write file show speed/written

AbsoluteCatalyst:
I added the ability for us to pass a newline character to waitScreen and for it to split the text onto the next row. Works with PSTR too. 
SD file writing now displays a busy status that shows the current written bytes and the current write speed. 
I tried keeping it all fairly lightweight.


repetier:
Good idea for longer messages or with 2 different facts.

PR #1060, opened 2021-02-04
GUI - Allow Compiling-out encoder repeat features

AbsoluteCatalyst:
Setting ENCODER_MAX_REPEAT_STEPS to 0 should compile out the encoder configuration stuff.


repetier:
Thanks for adding this. My count was 71 bytes for AVR we can reduce.

PR #1061, opened 2021-02-04
STM32F1 (Mini E3) - Add flto flag & enable dedicated SPI 

AbsoluteCatalyst:
Turning on -flto for STM32F1 since the E3's flash is getting a *little* tight with all features turned on.

PS. It turns out AVR has flto turned on by default now in PIO. 
Turning it off via unflag can speed up compile time a fair bit (if you have the RAM/flash to spare.)


repetier:
> PS. It turns out AVR has flto turned on by default now in PIO.
Thanks, explains why enabling it did not make a difference is size. But since ram is always tight on AVR I keep it then enabled.

PR #1063, opened 2021-02-04
AVR - RAM reductions 

AbsoluteCatalyst:
Two quick config changes for AVR. 

Added macro conditions to disable I2C for AVR if unused. Saves us around 143 bytes. 
Just have to make sure Wire.h isn't included anywhere. We can use the macro "NO_I2C" to disable it.

I switched the U8G2 display page buffer mode for AVR to mode 1 (slowest) but uses a lot less RAM (128 bytes)


repetier:
Good ideas on reduction. Looks like we can make avr a full functional member for most use cases after all. User just has to make sure he is not using greedy modules:-)

AbsoluteCatalyst:
Aye. 
I think there's a few more tiny tiny code related changes we could do, but do you think 1600~ bytes free is enough? 
Have you been able to compile using the u8g2 display yet?

repetier:
I have it currently running with 800 byte free code. Hope there is no path where local variables get more:-)
I know I could change serial handling which currently doubles buffer memory for emergency parser solution.

Currently I have it on a 20x4 display because that is what the test printer has. The only active printer with AVR currently. 32 bit is already wide spread or are commercial printers with special stuff like Prusa where firmware is not compatible at the moment.

repetier:
When I'm through I will test it with a plain board and AVR.

PR #1072, opened 2021-03-27
Method to check if controller has not been used for a period of time

nukem:
The method can be used to turn on and off lights like controller backlight and case lights.

```
#if defined(UI_BACKLIGHT_PIN) && UI_BACKLIGHT_PIN > 0
IO_OUTPUT(controllerBackLightPin, UI_BACKLIGHT_PIN)
LIGHT_STATE_MONOCHROME(controllerBackLightState)
**LIGHT_COND(controllerBackLightState, true, !GUI::isGUIIdle(), 255, 255, 255, 255)**
LIGHT_SOURCE_MONOCHROME(controllerBackLightDriver, controllerBackLightPin, controllerBackLightState)
#endif

IO_OUTPUT(caseLightPin, CASE_LIGHT_PIN)
LIGHT_STATE_MONOCHROME(caseLightState)
LIGHT_COND(caseLightState, true, Printer::caseLightMode, 255, 255, 255, 255)
LIGHT_COND(caseLightState, Printer::isUIErrorMessage(), LIGHT_STATE_BLINK_SLOW, 255, 255, 255, 255)
LIGHT_COND(caseLightState, Printer::isDoorOpen(), LIGHT_STATE_ON, 255, 255, 255, 255)
**LIGHT_COND(caseLightState, true,  !GUI::isGUIIdle(), 255, 255, 255, 255)**
LIGHT_SOURCE_MONOCHROME(caseLightDriver, caseLightPin, caseLightState)

```

### DIFF
--- a/src/Repetier/src/controller/gui.cpp
+++ b/src/Repetier/src/controller/gui.cpp
@@ -14,6 +14,7 @@ GUIStatusLevel GUI::statusLevel = GUIStatusLevel::REGULAR;
 bool GUI::contentChanged = false;                                 ///< set to true if forced refresh is wanted
 GUIAction GUI::nextAction = GUIAction::NONE;                      ///< Next action to execute on opdate
 int GUI::nextActionRepeat = 0;                                    ///< Increment for next/previous
+uint8_t GUI::idle = 0;
 
 #if ENCODER_MAX_REPEAT_STEPS != 0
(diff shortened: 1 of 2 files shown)


repetier:
No need to wast memory with a state. I have uploaded an alternative where the computation is done in GUI::isGUIIdle() directly. The few extra cycles there should be no problem.

nukem:
Ah ok, I thought it would hang if it was inline.

PR #1074, opened 2021-03-29
Fix wrong reference to NUM_BEEPERS

nukem:


repetier:
Bad type. Explains why it did not work for you also it should, thanks.

PR #1084, opened 2022-05-01
Allow space in commands M23 / M30 /  M32 / M118

luc-github:
M20 list long name but command using this  if have a space are not working:
so this PR fix it for M23 / M30 / M32

M118 like M117 can have also space in it so sync with M117 

I have updated AVR/ DUE code- I hope is ok  - long time I did not do PR here ^_^ - let me know if I need to change something

I did not tested dev2 branch - as It seems I do not have supported board I have AVR which seems not yet ported in Readme, and my Due is dead sorry


repetier:
Thanks for the patch. Also untested i can see that spaces would break these commands. For V2 I did change it my self manually as it lacks same problem.

PR #1086, opened 2022-08-02
Fix build on macOS and don't fail with M21 silently

arekm:
1)
    Fix build on macOS due to unwanted characters.

    Building fails with:
````
    Eeprom.cpp:1:1: error: stray '\357' in program
     /*
     ^
    Eeprom.cpp:1:2: error: stray '\273' in program
     /*
      ^
    Eeprom.cpp:1:3: error: stray '\277' in program
     /*
       ^
````

2)  M21 command (sd card init) fails silently.

When there is no sd card then M21 just fails silently:

````
    Send: N3 M21*19
    Recv: ok 3
````

which confuses octoprint. Since no failure was reported octoprint thinks
sd card is supported while in reality it is not.

With this patch failure is properly signalled:

````
    Send: N9 M21*25
    Recv: ok 9
    Recv: SD init fail
````

When sd card is inserted and M21 is issued:

````
    Send: N10 M21*33
    Recv: ok 10
    Recv: Card successfully initialized.
````

Happens on RAMPS 1.4 board.

### DIFF
--- a/src/ArduinoAVR/Repetier/Eeprom.cpp
+++ b/src/ArduinoAVR/Repetier/Eeprom.cpp
@@ -1,4 +1,4 @@
-﻿/*
+/*
     This file is part of Repetier-Firmware.
 
     Repetier-Firmware is free software: you can redistribute it and/or modify
--- a/src/ArduinoAVR/Repetier/SDCard.cpp
+++ b/src/ArduinoAVR/Repetier/SDCard.cpp
@@ -71,8 +71,10 @@ void SDCard::initsd() {
     sdactive = false;
 #if SDSS > -1
 #if SDCARDDETECT > -1
-    if (READ(SDCARDDETECT) != SDCARDDETECTINVERTED)
+    if (READ(SDCARDDETECT) != SDCARDDETECTINVERTED) {
+        Com::printFLN(Com::tSDInitFail);
         return;
+    }
 #endif
     HAL::pingWatchdog();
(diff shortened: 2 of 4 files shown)


repetier:
Thanks. Have updated github and also the current dev version.

But shouldn't you not better watch for
"Card successfully initialized."
to tell octoprint that it is supported and working? Would also solve problem with old. Or use both - start undecided and if you either get an error or success you are sure what is the state.

Since you are the sd expert and I'm currently reworking sd handling in V2 firmware to support up to 4 solutions in parallel - is there a solution you know about to tell hosts how many different sd/usb storage devices are possible in parallel and how to select them? I saw BTT displays use Px for it. But RepRapFirmware also uses P just as string. But you need to know the names. And I haven't seen any description telling that they do.

As fallback I plan to use first inserted device in fixed order so as long as only one device is connected it would work as always.

arekm:
octoprint works in "stard undecided" way but assumes sd card is there initially. From what I see it doesn't handle "Card successfully initialized" but only "SD card ok" and since it assumes sd card is there only failure message can tell it there is no sd card. Edit: it's more complicated - initial state is a user changeable setting!

More user than expert.

I see M408 S1 reporting nr of slots ("volumes"). M39 for reporting sd card status for n slot (in json or as text depending on S parameter). ReprapFw M21/22 Px for selecting sd card slot number.

repetier:
Yes, that is RepRap style solution as it uses json compared to us or marlin. So most likely will be a new special case:-( But I try to stay close so fallback should work like old ones plus some ways for more if supported.

PR #1088, opened 2022-08-20
Create uilang.h

Sakgoo:
NOTE: The Turkish translation part includes corrections. In this file, the deficiencies in the Turkish part have been corrected. Translated into Turkish 98%. I wanted to make a small contribution. I wish you good work. MD...


repetier:
Not sure what is wrong, but there is only one uilang.h which seems completely new. Also request is against master and not dev.

In dev I recently updated the uilang.h file (for 1.0.5) with turkish translations. Are these the same changes? In that case we can just skip it, as I already have added it to 1.0.5/dev branch.

Sakgoo:
Hello. I'm just reading your comment. I am sorry for this. I was out of the office. The update I sent was newer. but it was for version 1.0.4. Now I am submitting the updated uilang.h file for version 1.0.5. Good work.

